PR #27, opened 2021-11-07
Fix "chip_icon_only", "chip_icon_state" and "chip_icon_double_state"

CM000n:
Since the basic chip template does not have an icon section, no dependent chip cards can use it. Due to this MR "`chip_icon_only`", "`chip_icon_state`" and "`chip_icon_double_state`" use variables that are only mapped via the template section.

### DIFF
--- a/config/minimalist-templates/button_card_templates.yaml
+++ b/config/minimalist-templates/button_card_templates.yaml
@@ -60,7 +60,7 @@ chip_icon_state:
       if (variables.ulm_chip_icon_state_icon){
         var icon = variables.ulm_chip_icon_state_icon;
       } 
-      var state = 'default';
+      var state = '';
       if (states[variables.ulm_chip_icon_state_entity].state){
         var state = states[variables.ulm_chip_icon_state_entity].state;
       } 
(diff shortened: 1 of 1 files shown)


Paddy0174 (commented):
Thanks for the PR. Is there any specific reason, why you are declaring the variable before the if-statement and again inside? 
```yaml
var icon = '❔'; // here
if (variables.ulm_chip_icon_only){
  var icon = variables.ulm_chip_icon_only; // and here
```
If not, could you please change that, so I can merge the PR. Thanks!

CM000n:
Hi @Paddy0174. I followed the logic of the previous Chips templates, which also used a '❔' placeholder icon if nothing else was specified in `lovelace-minimalist.yaml`.

By specifying the variable in advance, it is set in any case and will be used if the check `if (variables.ulm_chip_icon_only)` returns `false`. Otherwise, the icon/emoji from `variables.ulm_chip_icon_only` is used, which is specified within the if function.

schumijo:
Hi @CM000n,

Be carreful, in your commit there are some typos  : 
triggers_update:
    - "[[[ variables.ulm_chip_icon_state_entity_1 ]]]" <==== Should be variables.ulm_chip_icon_double_state_entity_1
    - "[[[ variables.ulm_chip_icon_state_entity_2 ]]]" <==== Should be variables.ulm_chip_icon_double_state_entity_2
   
var state = states[variables.ulm_chip_icon_double_state_entity_1].state; <===  Should be var state1
var state = states[variables.ulm_chip_icon_double_state_entity_2].state; <===  Should be var state2

Also, I think it could be cool to add unit_of measurement for each entities

CM000n:
I haven't changed anything in the trigger variables ;-) But you are absolutely right.
Thanks, I will fix that

Paddy0174:
> 
> 
> Hi @Paddy0174. I followed the logic of the previous Chips templates, which also used a '❔' placeholder icon if nothing else was specified in `lovelace-minimalist.yaml`.
> 
> By specifying the variable in advance, it is set in any case and will be used if the check `if (variables.ulm_chip_icon_only)` returns `false`. Otherwise, the icon/emoji from `variables.ulm_chip_icon_only` is used, which is specified within the if function.

That's totally right, but wasn't what I meant. :) I meant that you're declaring the variable twice. Before the if-statement you use "var" and inside again, where inside the "var" should be ommitted. You're not changing the value of the variable, you're declaring it again with a new value. :) But no big deal, really! :+1: If you don't mind, I'll just change it. 

And the wrong variables for `triggers_update` are/were wrong from my side. But thanks, this is fixed as well with your PR! So you're actually solving two problems at once! 👍  🤣

PR #34, opened 2021-11-10
Add custom card custom_card_saxel_fan

saxel:
Added my version of a fan card that allows for setting the fan speed with a slider, and oscillate toggle using a button. Also added a _Blue_ style inspired by 7ahang's work on Behance.

light_theme
dark_theme


CM000n:
Hi. @TBens, has now given me the right for the UI repo. Unfortunately, I don't have the rights for the page repo to revise the wiki yet.
I would still include this card in the next release and close this PR here. I'll then make a separate bug in the page repo so we can add to the wiki once someone has permission.

Issue #33, opened 2021-11-09
Option to use MDI icons in chips


basbruss:
The chips don't show the MDI icons when using the variables.
Only when using emoji's
```
- type: 'custom:button-card'
  template: chip_icon_state
  entity: group.denon
  variables: 
    ulm_chip_icon_state_icon: 'mdi:speaker-wireless'
    ulm_chip_icon_state_entity: sensor.denon_count
```
<img width="509" alt="Screen Shot 2021-11-09 at 11 32 40" src="https://user-images.githubusercontent.com/68892092/140908605-8d09cc83-c537-4266-b8cd-d08d39b48c38.png">
.

Paddy0174:
That is intentional. :) The chips are, theoretically, only used in combination with emojis. But if there's a need for mdi-icons instead of emojis, I can take a look and see what we can come up with. Noted and on the to-do-list, but not on highest priority. 

I'll change the explanation in the wiki to reflect this issue and point out the differences. ;) 

On another task, where did you get these pictures for the person card? Is that kind of a generator or did you do these with Photoshop or something alike? Are you using this with dutch language? Would you mind sharing your language file, so we can include it in the main repo? Thanks!

Paddy0174:
I just took a quick look, this should be doable, but it needs some "bigger" changes on the code. 

For the time being, this would be a quick and dirty solution for **your** layout. Not tested, just from the top of my head.
This is what you should have now:
```yaml
type: 'custom:button-card'
template: chip_icon_only
variables:
  ulm_chip_icon_only: '❤️'"
```
Try it with this for every `chip_icon_only` you want an mdi-icon to appear:
```yaml
type: 'custom:button-card'
template: chip_icon_only
variables:
  ulm_chip_icon_only: null
icon: 'mdi:face-man'
label: null
```
As I said, not a very clean way, but it should do the trick for now. :) Hope this helps a bit!

basbrussee:
> That is intentional. :) The chips are, theoretically, only used in combination with emojis. But if there's a need for mdi-icons instead of emojis, I can take a look and see what we can come up with. Noted and on the to-do-list, but not on highest priority.
> 
> I'll change the explanation in the wiki to reflect this issue and point out the differences. ;)
> 
> On another task, where did you get these pictures for the person card? Is that kind of a generator or did you do these with Photoshop or something alike? Are you using this with dutch language? Would you mind sharing your language file, so we can include it in the main repo? Thanks!

[...]

PR #38, opened 2021-11-12
Add custom single icon chip card

CM000n:
Hi guys, 
I created a custom chip card today that can handle MDI icons and looks at more information on a tap.
The official chip icon cards available so far can only handle emojis. But unfortunately there are no decent emojis for home appliances ;-)
I thought maybe this chip card is still useful for others.
Greetings


Paddy0174:
Thanks, I hadn't had the time to work on that, but as #33 shows, there already is a demand! :) 👍 

Would it be ok for you, if I include this not as `custom`, but rather in the main template file as one of the default `chip`s? @CM000n please let me know, so I can include it in the next release (either way, just need to know which).

CM000n:
Of course, it would be an honor. I'll see that I can change the PR accordingly by tonight. 🙂

CM000n:
Hi @Paddy0174 , I just changed my PR accordingly. Hope that fits. If there is anything else I can do, let me know.

PR #39, opened 2021-11-15
Add custom card custom_card_schumijo_car

schumijo:
Added my version of a car card
car


Paddy0174:
I like the idea with the second dot on the icon. That should come in handy in various places. :)

PR #46, opened 2021-11-25
Round temperature when the temperature has decimal points

bavo:
Round temperature in the temperature chip card. 

Eg: 💧15.00° / 21.54° will be converted to 💧15.0° / 21.5°

Bonus: Removed all the trailing white spaces from the template file


bavo:
If there are conflicts with the merge due to all the white space changes, let me know. I will fix them ;)

PR #66, opened 2021-12-06
Vertical buttons custom state

saxel:
Added an extension of vertical_buttons template so the button can be active depending on the state of the entity. This is how TBens managed the "Scenes" by having an input_select, and when tapping the button card the service for selecting the option is called.

vertical_buttons_custom_state

| Variables | Description |
|-----------|-------------|
| state     | Compare the entity state value with this value, ex: the button will be **on** if the entity.state == state |
| color     | color for the when the button is **on**. The color names are defined in the theme. The following colors are supported: blue, green, yellow, pink, red, purple |

configuration.yaml
```yaml
input_select:
  status_person:
    name: 'Status • Person'
    options:
      - home
      - away
      - exercise
      - work
      - sleep
```

lovelace-minimalist.yaml
```yaml
- type: 'custom:button-card'
  template: vertical_buttons_custom_state
  variables:
    state: home
    color: purple
  entity: input_select.status_person
  icon: 'mdi:home-variant'
```

The new template is a replacement of the vertical_buttons_colorname, and removes the need for duplicating the code for each color, so I removed the templates `vertical_buttons_blue` and `vertical_buttons_green` as I didn't want to create new templates for the additional colors that we could have. But that also means it could break the dashboard if anyone was using those. Similarly I also removed `scene`, `scene_blue` and `scene_green`. Let me know if you have any concerns, or disagree with the approach.


KaherdinTristan:
At the risk of being really stupid, but, is this Github rep still being managed by anyone?

CM000n:
Unfortunately, this is not such a stupid question :( @Paddy0174 last activities were more than a month ago. I have also tried to write to him via the Home Assistant forum. So far, unfortunately, without feedback.
I really hope that he is well

saxel:
Yeah, unfortunately I don't know either and hope he's well too.

KaherdinTristan:
I too have written to him on the home assistant forum. No reply. Last seen approx 25 Nov. 
And i don't have his private email address either. 
🥺
Hope nothing has happened...

bavo:
It should be possible to continue to maintain the code base because paddy created an organization for this repo instead of managing it on his own. But I don't know who was added to the [Minimalist-UI](https://github.com/orgs/UI-Lovelace-Minimalist) org. 

But let's hope he's fine first of all. :crossed_fingers:

CM000n:
Unfortunately, if he hasn't added anyone to the organization, that won't do much good either. :-( Unfortunately, at least there are no public members: https://github.com/orgs/UI-Lovelace-Minimalist/people

KaherdinTristan:
Would one solution to this be that one with good knowledge, and maybe someone that we know has contributed to the projected forks this repo and immediately creates such an org for the FORK?

KaherdinTristan:
@schumijo ?

schumijo:
Hi
@TBens and @Paddy0174  are administrators of this repo. I am only a contributor. I also wait for a 'hello' from paddy...but nothing for many weeks 🤞
Let's wait for few more days before thinking to a fork. I have some knowledge about the code but don't have a lot of time and a lot of knowledge about GitHub 😁

CM000n:
No need for a fork. If @TBens realy is an administrator of this Organisation he can also give admin rights to someone from the contributors. But yeah, let's wait some more days.

CM000n:
H, like I wrote already in https://github.com/UI-Lovelace-Minimalist/UI/pull/34
@TBens, has now given me the right for the UI repo. Unfortunately, I don't have the rights for the page repo to revise the wiki yet.
I would still include this card in the next release and close this PR here. I'll then make a separate bug in the page repo so we can update the wiki once someone has permission.

saxel:
Sure, I'll look into it :)

willyp713:
@saxel since this change removed the option for scenes in the vertical buttons, how do we use a button to call a scene now?

I duplicated `card_vertical_buttons` to make a custom card, and added the following to _tap_action service_:

```
if( entity.entity_id.startsWith("scene.") )
          return "scene.turn_on";
```

But I really don't want to maintain a custom card just for that. Is there a better way to do this using the existing card?

mp-se:
The inteded wyay of using the cards is to use inputboolean or inputselect to choose/ativate the scene and then have an automation that triggers on the state change to start/end the scene. With inputboolean you can toggle a scene and with input select swtich between scenes. 

With an automation you can also store the current state of the entities and restore that afterwards. I can add support for the scene in the vertical button as well, I can submit a pr later today if you want to use it as you have before. Let me know.

willyp713:
> The inteded wyay of using the cards is to use inputboolean or inputselect to choose/ativate the scene and then have an automation that triggers on the state change to start/end the scene. With inputboolean you can toggle a scene and with input select swtich between scenes.
> 
> With an automation you can also store the current state of the entities and restore that afterwards. I can add support for the scene in the vertical button as well, I can submit a pr later today if you want to use it as you have before. Let me know.

Trying to understand here -- so I should create an input boolean helper, then have that input boolean activate an automation to turn on the scene? Right now I only have the scenes created, but it shouldn't be too much trouble to create the helper and the automation that is triggered by the helper.

Or should it be even easier than that?

Not sure we need a code change yet; this might just be a knowledge gap on my part.

mp-se:
Yes, thats the approach. The original vertical buttons only supported input_select to switch between scenes, we just added the input_boolean to make it toggle. I think I added support for switch and light to support other requests. so making it toggle a scene is not that hard. 

This is one of my automations that also create a snapshot of the entities which is restored when i untoggle the scene.
```
- id: '1640815805682'
  alias: Scene - Kitchen work
  description: ''
  trigger:
  - platform: state
    entity_id: input_boolean.scene_kitchen_work
  condition: []
  action:
  - choose:
    - conditions:
      - condition: state
        entity_id: input_boolean.scene_kitchen_work
        state: 'on'
      sequence:
      - service: scene.create
        data:
          snapshot_entities: light.kitchen_2
          scene_id: snapshot_kitchen_work
      - scene: scene.lights_kitchen_work
    default:
    - scene: scene.snapshot_kitchen_work
  mode: single
```

willyp713:
> Yes, thats the approach. The original vertical buttons only supported input_select to switch between scenes, we just added the input_boolean to make it toggle. I think I added support for switch and light to support other requests. so making it toggle a scene is not that hard.
> 
> This is one of my automations that also create a snapshot of the entities which is restored when i untoggle the scene.
> 
> ```
> - id: '1640815805682'
>   alias: Scene - Kitchen work
>   description: ''
>   trigger:
>   - platform: state
>     entity_id: input_boolean.scene_kitchen_work
>   condition: []
>   action:
>   - choose:
>     - conditions:
>       - condition: state
>         entity_id: input_boolean.scene_kitchen_work
>         state: 'on'
>       sequence:
>       - service: scene.create
>         data:
>           snapshot_entities: light.kitchen_2
>           scene_id: snapshot_kitchen_work
>       - scene: scene.lights_kitchen_work
>     default:
>     - scene: scene.snapshot_kitchen_work
>   mode: single
> ```

How does it go back to your stored snapshot after you toggle that `input_boolean` off? Is that just built-in behavior or do you have to define that in another automation?

willyp713:
Also for what it's worth I think I'm going with the `input_select` behavior which seems to automatically cycle between the scenes based on button presses. For example, I can set up several lighting scenes and control them as follows:

```
### INPUT SELECT ###
input_select:
  lighting_scene_select:
    name: lighting_scene_select
    options:
      - "scene.all_lights_on"
      - "scene.movie_mode"
      - "scene.sleep"
    icon: mdi:lightbulb-multiple-outline

### AUTOMATION ###
- id: '1643064813719'
  alias: Set Lighting Scene from Selection
  description: ''
  trigger:
  - platform: state
    entity_id: input_select.lighting_scene_select
  condition:
  - condition: template
    # check to confirm state is actually changing
    value_template: '{{ trigger.to_state != trigger.from_state }}'
  action:
  - service: scene.turn_on
    data_template:
      entity_id: '{{ states(''input_select.lighting_scene_select'') }}
  mode: single

### BUTTON CARD ###
- type: 'custom:button-card'
  template: card_vertical_button
  entity: input_select.lighting_scene_select
  name: Lights On
  variables:
    ulm_card_vertical_button_state: scene.all_lights_on
    ulm_card_vertical_button_color: yellow
```

mp-se:
> automation?

The default behavious is run when the button is off and this does the restore and runs the stored scene with the saved states

PR #67, opened 2021-12-10
Split button card templates

sildehoop:
I have splitted the button card templates.yaml file for convenience and scalability.


CM000n:
I don't know if I like the idea of a split config/template file. Even if it seems easier to maintain at first sight, you now have to know even more precisely where to look. Also, for users of the UI who are not technically proficient, it may be more cumbersome to have to copy multiple files. 
I leave the PR open for discussion

sildehoop:
Ok I do agree that it might be easier for the user to copy 1 file to maby we can find a middle way and compile to 1 file in the releases

sildehoop:
Then it's easier for the user and easier to mantain (I think)

saxel:
I just have a feeling it is solving a problem we don't have and I think the division of components are a bit too small/specific in some cases, and maybe it would be better to simply have a file for internal templates, and another one for card templates. Possibly a third one for popup cards depending on if it needs sub templates (depending on entity type) 🤷‍♂️.

With the current idea of the split (50+ new files, and more in future)  I don't think it is more convenient. VSCode outline gives me a good overview and allows me to quickly navigate in the file. But it is certainly more scalable if many people work like crazy with new templates that are still candidates as standard cards.

sildehoop:
Ok I understand, I thought if we start now we don't have to do it in the future. and I understand that it is not easy for the user because if a file is renamed then you get duplicate entries. for me is easier but that depends on your workflow.

sildehoop:
Only thing you can think about is that if more people work with this repository you will get more conflicts because everyone works in 1 file?

CM000n:
Well, the conflicts only arise when different changes are made to the same code at the same time. Changes to different code within a file are not really a problem for Github.
In addition, every developer should compare his code with the code base during longer development times in order to be able to address merge conflicts in good time.

Would it be ok for you if I close your PR for now?
Maybe we can go over the refactoring again at a later time. 
Otherwise we can discuss it here: https://github.com/UI-Lovelace-Minimalist/UI/discussions

PR #74, opened 2021-12-13
Add yaml linter and pre-commit hook to GitHub actions

CM000n:
Hello guys. I would like to suggest that in the future we use a linter to validate the yaml files. This will run automatically on pull requests or commits via a github action.

What do you guys think about this @schumijo, @saxel, @bavo?

I think this might be a bit more cumbersome for some of you during the initial development but I think in the long run we will be rewarded with more consistent code and won't be surprised by wrong indentations in .yaml files.

~~The linter itself does not perform any adjustments, but only issues warnings or errors, which must be adjusted by the respective developer. In addition, we could also think about having some carelessness errors (missing file extensions, trailing spaces, etc.) also automatically removed via pre-commit.~~

**..::edit::..**
I have now added a pre-commit workflow that now also checks the yaml files, among other things, and automatically makes some adjustments.


saxel:
Big thumbs up, I fully support the idea! 👍

schumijo:
Hi @CM000n 

You request my review because I made some PR on dev branch ? 
Or you need a "real" review of your change ? (to be honest I have no knowledge about linter but if it can help keeping a clean code, GO ! 👍  )

CM000n:
I was more interested in getting the opinion of the main contributors @schumijo, as it may influence your work the most.
The code itself I would still test myself. If this fits so, I would then push.
Possibly there are merge conflicts in your current PR's which would then have to be fixed by you.

bavo:
@CM000n I don't know if you already seen this page. But there is a default yaml style for home assistant. 
More info can be found here: https://developers.home-assistant.io/docs/documenting/yaml-style-guide

bavo (approved):

bavo:
Changes look good 👍 

The only thing I don't understand is why the 3 dashes at the start of every file? 
I'm not against it, but I just don't understand the reason 😛

CM000n:
The three dashes are not necessary, but a general yaml style guideline to separate directives from documents start.
https://newbedev.com/why-3-dashes-hyphen-in-yaml-file

CM000n:
I have now adapted it to the Homeassistant YAML Style Guideline (THX @bavo!). I tested the adapted code once, as far as possible, at home in my DEV Homeassistant. So far everything seems to work. I would therefore merge once, so you can continue to work on your pull request.

PR #80, opened 2021-12-14
More-info function light card and Custom auto-color light cards

basbruss:
Includes icon as more-info button for standard non-auto-color light card and different auto-color light card with icon as more-info button.

Dark mode:
Screen Shot 2021-12-14 at 20 10 11

Light mode:
Screen Shot 2021-12-14 at 20 19 02


CM000n:
Wow @basbruss , this is fantastic work! Thank you for your contribution!
I would have a few small comments though. 
It looks like your folder structure is wrong. Custom cards should always be under the `custom_cards` folder. However, in your case, the `custom_card_light_color` folder is directly in the root directory.
I would also suggest combining your individual .yaml templates into one file as much as possible. This makes copying the necessary code easier for the less technical user and is the .yaml template style we maintain here, at least so far.

But thanks again, really great job

basbruss:
Thanks! I have uploaded only the `custom_card_light_color` folder which is in the `custom_cards` folder on my system, to not upload all the other custom cards (which can potentially lead to conflicts in merging). I can split the folder in three files if this makes it easier to implement: `internal_templates.yaml`, `custom_light_cards_color.yaml` and `custom_light_cards_icon_info.yaml` (without the auto-coloring feature). To make a clear distinguish between features.
I am still working on the icon_info function for the other standard_light cards. I hope to add this in the coming week.
I agree that these cards need some documentation, esp. with light and dark mode switching. Because some JS-templates are based on the sun.state as HA does not provide for now a way to determine in JS if your system is in Dark or Light mode.
I am not that familiar with making a readme file. But I will take one as template from another custom_card to provide the necessarily documentation. :-)

CM000n:
Please look again. In your merge request the `custom_card_light_color` folder is definitely not inside the `custom_cards` folder 😉
wrong_folder

Regarding your other comment: Shouldn't we create two custom cards then? One for `custom_light_cards_color` and one for `custom_light_cards_icon_info`? That would make the structure clearer and you could put the internal templates into the respective custom card template files. This way we would have only one file per custom card.

I am eagerly awaiting your PR on the icon of the standard light card ☺️. Thanks a lot for your effort.

CM000n:
Thank you @basbruss . It looks much better now :-) 
It looks like the merge request checks are still failing because your .yaml files don't match our .yaml style guidelines.
If you want, I can send a PR to your repo later to fix the errors. Otherwise you can also have a look at the corresponding error message of the check yourself:
https://github.com/UI-Lovelace-Minimalist/UI/runs/4532476213?check_suite_focus=true

basbruss:
Oh wow! Didn't notice that. Uploaded it in unintentionally the main map. The branch is now updated with the suggested folder structure in custom_cards folder :)
I am still learning to use Git in a favorable way. Could you check what did the check making fail?

CM000n:
Most error messages come from not enclosing the string values in your .yaml files in double quotes:
error_messages

For example, this...:

```
card_light_color:
  template:
    - icon_info_light
    - ulm_language_variables
  variables:
    ulm_card_light_name: "[[[ return entity.attributes.friendly_name ]]]"
    ulm_card_light_icon: 
  show_icon: false
  show_name: false
  show_label: false
  state:
    - operator: template
      value: "[[[ return entity.state == 'on' ]]]"
      styles:
        card:
          - background-color: >
              [[[
                if (states['sun.sun'].state == "below_horizon"){
                  var color = entity.attributes.rgb_color;
                  if (color){
                     return 'rgba(' + entity.attributes.rgb_color + ',0.1)'
                  }
                  else{
                    return 'rgba(var(--color-yellow),0.1)'
                  } 
                }else if (states['sun.sun'].state == "above_horizon")
                  return 'rgba(var(--color-background-yellow),var(--opacity-bg))';
              ]]]
  styles:
    card:
      - border-radius: var(--border-radius)
      - box-shadow: var(--box-shadow)
      - padding: 12px
  tap_action:
    action: toggle

...Rest of your yaml file
```

... should look more like this:

```
---
card_light_color:
  template:
    - "icon_info_light"
    - "ulm_language_variables"
  variables:
    ulm_card_light_name: "[[[ return entity.attributes.friendly_name ]]]"
    ulm_card_light_icon: 
  show_icon: false
  show_name: false
  show_label: false
  state:
    - operator: "template"
      value: "[[[ return entity.state == 'on' ]]]"
      styles:
        card:
          - background-color: >
              [[[
                if (states['sun.sun'].state == "below_horizon"){
                  var color = entity.attributes.rgb_color;
                  if (color){
                     return 'rgba(' + entity.attributes.rgb_color + ',0.1)'
                  }
                  else{
                    return 'rgba(var(--color-yellow),0.1)'
                  } 
                }else if (states['sun.sun'].state == "above_horizon")
                  return 'rgba(var(--color-background-yellow),var(--opacity-bg))';
              ]]]
  styles:
    card:
      - border-radius: "var(--border-radius)"
      - box-shadow: "var(--box-shadow)"
      - padding: "12px"
  tap_action:
    action: "toggle"

...Rest of your yaml file
```
Hope you get the idea 😉

basbruss:
Okay, I formatted the code from the style errors. The only error still showing is `Error: The process '/opt/hostedtoolcache/Python/3.8.12/x64/bin/pre-commit'`. Do you know what is causing this?

CM000n:
Yes, there are still some flaws in your code:
error4

- You have some trailing whitespaces
- Your files don't have a proper file ending (The last line must be empty and must not contain any code)
- And you have still some string values not correctly quotet

For example this... 
`      - grid-template-areas: '"item1" "item2"'`

...should be quotet like this:
`      - grid-template-areas: "'item1' 'item2'"`

CM000n:
Yeah! All checks passed. Great work! :-)

basbruss:
I've also added the other normal light card with the icon as more-info button :)
So I only need to make the readme files when a have some spare time.

CM000n:
Great 🥰
What is your opinion on this. Do you want to create a separate merge request for the documentation or should I leave this one open long enough for you to add the documentation to it?

basbruss:
Hi, I've added for both cards the readme files. So I think the cards are ready to merge :)

PR #85, opened 2021-12-15
Add rounding and german translation to schumijo car card

CM000n:
Hello @schumijo , I would like to suggest removing the decimal places for the energy level and range on your custom_card_schumijo_car. 
I think the card is great and I use it myself. However, my Toyota integration provides up to 6 decimal places for the fuel level. This is, in my opinion, nice to have but unnecessary in the UI. There, a glance at the value without decimal places is actually enough for me to inform myself briefly.

BTW: I have also added a German translation :-)


schumijo (approved):
Approved

PR #103, opened 2021-12-17
Fix guest presence counter

checkerschaf:
This fixes the guest presence counter by making the field optional.

Before you had to set `ulm_chip_presence_counter_guests = ""` to make it optional what seems counter intuitive to me.
I guess this should make it more easy to use.


CM000n:
Määäh! The sheep has checked!
Thank you very much. This is indeed a better approach

PR #114, opened 2021-12-20
Make card_person zones optional and more robust

checkerschaf:
This fixes some errors when users input wrong data into the zones.
Additionally this makes them optional as some users may not have two zones and therefore the card won't display at all when the user isn't at `home`.

I added some safe equal operations as some of the data could potentially be `""` or a boolean 😄


CM000n:
Määäh! Cool, thank you very much

PR #122, opened 2021-12-21
New Thermostat Card

mlittle408:
Here's a new thermostat card based on @TPX01 air-condition. This new card supports heat, cool and heat_cool hvac_modes.


CM000n:
Nice! Thank you very much @mlittle408. I'm just wondering if it wouldn't be better to customize the existing card instead of adding a new custom card? Not that we don't end up with several thermostat cards that look the same and do the same thing, but just a little differently.

mlittle408:
Sure that's up to you. No problem.

CM000n:
lol, what was that? 😀 I actually thought we could discuss this here and work on it together.

mlittle408:
Oh, I apologize. I'm pretty flexible. I'll do it however you like.

PR #124, opened 2021-12-21
IT Language for Welcome card.yaml

httpedo13:
Add Italian language.


CM000n:
Very cool! Thanks for the translations @httpedo13

PR #153, opened 2021-12-26
Fix battery_card icon color when state is unavailable

saxel:
The battery card showed the icon as green when the state is unavailable.  I think it would be more appropriate to show the default color in such case.

Before:
image

After (Light Mode):
image

After (Dark Mode):
image


CM000n:
Oops, something went wrong here. I accidentally deleted the dev branch. But wait with a new PR. Maybe we move the repo this week

PR #156, opened 2021-12-27
Fix custom_card_saxel blue style card

saxel:
After the commit "Add pre-commit hook and modify files" my blue style custom card was not formatted properly due to this change. I think it was a simple mistake.


CM000n:
Oops, something went wrong here. I accidentally deleted the dev branch. But wait with a new PR. Maybe we move the repo this week

PR #162, opened 2021-12-27
Use new slider from AnthonMS for popup card

bavo:
Fix using 2 different sliders in the light popup. 

see https://community.home-assistant.io/t/lovelace-ui-minimalist/322687/872?u=bms


bavo:
Added new parameter to hide all the color wheels for the child entities within a light group

PR #186, opened 2022-01-02
Make state of person card translatable

pierrecle:
By default, person card label is the state, which have two default values, `home` and `away`. We can quiet easily translate these states for a nicer card :)


config/minimalist-templates/button_card_templates.yaml
```diff
@@ -1048,6 +1054,8 @@ card_person:
               return "stretch stretch";
             }
           ]]]
+    label:
+      - text-transform: "capitalize"
```
CM000n, line 1058: Thank you, that is a really very good idea to make the labels of the person cards translatable. I just don't know yet whether I like the idea that the label is now printed in capital letters. It doesn't really fit in with the way it is displayed everywhere else. What do the rest of you think?

pierrecle:
In fact, it can be left in lower case, I'll change it in variables/style if I want to :)

PR #202, opened 2022-01-04
New sun card

httpedo13:
Hi, I have adapted the [sun card](https://github.com/AitorDB/home-assistant-sun-card)for minimalist ui.


schumijo:
Hi @httpedo13,

Good job...but I don't understand why creating a custom card 🤔
Maybe we could just write a "design guideline" for that kind of things ?
Just adding this CSS property with card_mod will produce the same result without using button-card ?
```
border-radius: "var(--border-radius)"
box-shadow: "var(--box-shadow)"
padding: "12px"
```

CM000n:
> Hi @httpedo13,
> 
> Good job...but I don't understand why creating a custom card 🤔 Maybe we could just write a "design guideline" for that kind of things ? Just adding this CSS property with card_mod will produce the same result without using button-card ?
> 
> ```
> border-radius: "var(--border-radius)"
> box-shadow: "var(--box-shadow)"
> padding: "12px"
> ```

Even if I am basically with you @schumijo, imho it is still easier for less technically oriented users to copy a custom card template than to deal with a manual specification of car-mod attributes.
Also it makes the workflow and the dashboard config more consistent if only custom cards are used.
So yes, I think it's perfectly fine to make a custom card here. In addition, we can certainly still think about an extension of the style guidelines :-)

httpedo13:
In my opinion a button card is accessible to everyone.

bavo:
It certainly is more consistent with the current cards we have 😉

schumijo:
Ok guys 😉

PR #209, opened 2022-01-06
Fix black and white weather icons when bold

telashton:
Due to a bug in Chrome, emojis would lose colour if the font-weight was set to "bold". This change fixes this by wrapping the icon in a span and explcitly giving it a weight of 500 to preserve colour, leaving the label bold.

### DIFF
--- a/config/minimalist-templates/button_card_templates.yaml
+++ b/config/minimalist-templates/button_card_templates.yaml
@@ -44,7 +44,7 @@ chip_temperature:
       }
       var outside_temp = states[variables.ulm_chip_temperature_outside].state;
       var inside_temp = states[variables.ulm_chip_temperature_inside].state;
-      return "<span style='font-weight:500;'>" + icon + "</span>" + " " + convertTemperature(outside_temp) + "° / " + convertTemperature(inside_temp) + "°" ;
+      return icon + " " + convertTemperature(outside_temp) + "° / " + convertTemperature(inside_temp) + "°" ;
     ]]]
 chip_icon_only:
   template: "chips"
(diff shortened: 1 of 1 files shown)


config/minimalist-templates/button_card_templates.yaml
```diff
@@ -1361,7 +1361,7 @@ chips:
     label:
       - justify-self: "center"
       - padding: "0px 6px"
-      - font-weight: 500
```
CM000n, line 1364: By setting the font weight in the chip base template to 500, we have ensured that the emojis are displayed correctly for all chips.
With your solution, we would now have to take care of each chip individually, wouldn't we?

telashton: While setting the font-weight in the base template does indeed fix the colourisation of the emojis, it reduces the text weight and therefore deviates from tben's original design. The chips_icon_label template (which is used in chips_alarm) has the correct weight weight and if you have this side-by-side with another chip the difference is quite noticeable. In CSS, the numerical equivalent to bold is 700, rather than 500.

To fix for all chips that use an emoji in the label, we would need to replace:

`return icon`

with

`"<span style='font-weight:500;'>" + icon + "</span>"`

in each template and set the font-weight to bold in the chip base template. My personal setup only utilises the temperature chip, which is reflected in the PR.

CM000n: It would be cool if you could make a more general solution in your PR. Since the black and white icons are a bug in the Chromium engine and not actually in the theme, it should be seen as a temporary workaround. We would hate to treat each card/chip individually.

telashton: Looks like this bug has been resovled in v97 of Chrome, so only change required is reverting the font-weight to bold from 500 in the base chip template.

CM000n:
Ahh, I didn't realize that, thank you. I will have a look.

Issue #191, opened 2022-01-03
Enable "switch mode" for vertical_buttons_custom_state


arndth:
I was trying to implement vertical buttons in my environment and it did not work for me. The color were not shown and when got the color it still was not changing the state from 'on' to 'off'. 

I changed the default state and the tap on action. In future a flexible color as variable would also be helpful for the community. Here my code: (see also post in [Forum](https://community.home-assistant.io/t/lovelace-ui-minimalist/322687/938?u=arndth))

```yaml
  vertical_buttons_custom_state:
      template: "vertical_buttons"
      variables:
        state: "on"
        # blue, red, green, yellow, purple, pink
        color: "green"
      show_last_changed: false
      name: "[[[ return variables.state ]]]"
      styles:
        name:
          - text-transform: "capitalize"
      state:
        - operator: "template"
          value: "[[[ return entity.state == variables.state ]]]"
          styles:
            icon:
              - color: "[[[ return `rgba(var(--color-${variables.color}), 1)`; ]]]"
            label:
              - color: "[[[ return `rgba(var(--color-${variables.color}-text), 1)`; ]]]"
            name:
              - color: "[[[ return `rgba(var(--color-${variables.color}-text), 1)`; ]]]"
            img_cell:
              - background-color: "[[[ return `rgba(var(--color-${variables.color}), 0.2)`; ]]]"
            card:
              - background-color: "[[[ return `rgba(var(--color-background-${variables.color}), var(--opacity-bg))`; ]]]"
      tap_action:
        action: "call-service"
        service: "input_select.select_option"
        service_data:
          entity_id: "[[[ return entity.entity_id ]]]"
          option: >
            [[[
              if (entity.state == 'on') {
                return 'off';
              } else if (entity.state == 'off') {
                return 'on';
              }
            ]]]
```

CM000n:
I think you are using the card template wrong. 
You have tinkered with our internal template, where the state and color can already be specified as variables in the templates intended for end users.
Please have a look at the PR for the map from @saxel: https://github.com/UI-Lovelace-Minimalist/UI/pull/66

_So the end users template to use for this card would for example be rather:_

configuration.yaml

```
input_select:
  status_person:
    name: 'Status • Person'
    options:
      - home
      - away
      - exercise
      - work
      - sleep
```
     
lovelace-minimalist.yaml

```
- type: 'custom:button-card'
  template: vertical_buttons_custom_state
  variables:
    state: home
    color: purple
[...]

Issue #198, opened 2022-01-03
Icon Only Light/Switch


damianeickhoff:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
I Would like a button-only light switch. I made a simple one which can turn on and off the light ór switch and also gives it the yellow color when on. But, it doens't show a warning if a light is unavailable. When I add that template, the padding gets messed up.

Also the padding itself doenst fit the whole theme. I hope maybe someone can make this better then me :)

**Additional context**
![afbeelding](https://user-images.githubusercontent.com/12126826/147946920-056f7391-0cc9-4fca-a20b-6e98560b6933.png)

Code:
```
icon_only_custom:
  color: "var(--google-grey)"
  show_icon: true
  show_label: false
  show_name: false
  template:
    - "yellow"
  tap_action:
    action: "toggle"
  hold_action:
    action: "more-info"
  styles:
    icon:
      - color: "rgba(var(--color-theme),0.2)"
    label:
      - justify-self: "start"
      - align-self: "start"
      - font-weight: "bold"
      - font-size: "12px"
      - filter: "opacity(40%)"
      - margin-left: "12px"
    img_cell:
      - background-color: "rgba(var(--color-theme),0.05)"
      - border-radius: "50%"
      - place-self: "center"
      - width: "42px"
      - height: "42px"
    card:
      - border-radius: "var(--border-radius)"
      - box-shadow: "var(--box-shadow)"
      - padding: "12px"
      - width: "84px"
  size: "20px"
```

sildehoop:
I will pick this up after the split merge

saxel:
Maybe this should be handled by #211 if it is also extended to support other entities than input_select and input_boolean. You get the desired look by specifying show_name: false

Example:
```yaml
              - type: 'custom:button-card'
                template: card_vertical_button
                entity: fan.fan
                show_name: false
                variables:
                  ulm_card_vertical_button_color: blue
```

![image](https://user-images.githubusercontent.com/12456858/148715772-288dfcd2-67d6-4653-90e1-f01f2f662998.png)

CM000n:
Should be added with https://github.com/UI-Lovelace-Minimalist/UI/pull/211

PR #211, opened 2022-01-06
Replacement card for vertical_button + vertical_button_custom_state

mp-se:
Adress feature request #191 , enable vertical button to be used as switch with input_booelan. Everything is now in one card which can be extended to handle other entites if needed. 

Tested the card with both input_boolean and input_select entities. 

I still have an issue with backwards compatibility parts for vertical_button_custom_state which i cant find the solution to. Need help on that one. More info in the feature request.

### DIFF
--- a/config/minimalist-templates/button_card_templates.yaml
+++ b/config/minimalist-templates/button_card_templates.yaml
@@ -279,20 +279,8 @@ card_title:
 ### VERTICAL BUTTONS (fka SCENES) ###
 card_vertical_button:
   variables:
-    ulm_card_vertical_button_color: |
-      [[[
-        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
-          if(typeof variables.color !== 'undefined')
-            return variables.color;
-         return "blue";
-      ]]]
-    ulm_card_vertical_button_state: |
-      [[[
-        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
-          if(typeof variables.color !== 'undefined')
-            return variables.state;
-         return "on";
-      ]]]
+    ulm_card_vertical_button_color: "blue"
+    ulm_card_vertical_button_state: "on"
   show_label: true
   label: ""
   name: |
@@ -355,6 +343,10 @@ card_vertical_button:
           return "input_select.select_option";
         if( entity.entity_id.startsWith("input_boolean.") )
           return "input_boolean.toggle";
+        if( entity.entity_id.startsWith("switch.") )
+          return "switch.toggle";
+        if( entity.entity_id.startsWith("light.") )
+          return "light.toggle";
         // If we need to support other entities we can add these options here.
         return "";
       ]]]
@@ -369,10 +361,67 @@ card_vertical_button:
       ]]]
 
 vertical_buttons:
-  template: "card_vertical_button"
+  show_label: true
+  label: "[[[ return (entity.attributes.value )]]]"
+  styles:
+    icon:
+      - color: "rgba(var(--color-theme),0.2)"
+    label:
+      - justify-self: "center"
+      - align-self: "start"
+      - font-weight: "bolder"
+      - font-size: "12px"
+      - filter: "opacity(40%)"
+    name:
+      - margin-top: "10px"
+      - justify-self: "center"
+      - font-weight: "bold"
+      - font-size: "14px"
+    img_cell:
+      - background-color: "rgba(var(--color-theme),0.05)"
+      - border-radius: "50%"
+      - place-self: "center"
+      - width: "42px"
+      - height: "42px"
+    grid:
+      - grid-template-areas: "'i' 'n' 'l'"
+    card:
+      - border-radius: "var(--border-radius)"
+      - box-shadow: "var(--box-shadow)"
+      - padding: "10px 0px 8px 0px"
+  size: "20px"
 
 vertical_buttons_custom_state:
-  template: "card_vertical_button"
+  template: "vertical_buttons"
+  variables:
+    state: "default"
+    # blue, red, green, yellow, purple, pink
+    color: "blue"
+  show_last_changed: false
+  name: "[[[ return variables.state ]]]"
+  styles:
+    name:
+      - text-transform: "capitalize"
+  state:
+    - operator: "template"
+      value: "[[[ return entity.state == variables.state ]]]"
+      styles:
+        icon:
+          - color: "[[[ return `rgba(var(--color-${variables.color}), 1)`; ]]]"
+        label:
+          - color: "[[[ return `rgba(var(--color-${variables.color}-text), 1)`; ]]]"
+        name:
+          - color: "[[[ return `rgba(var(--color-${variables.color}-text), 1)`; ]]]"
+        img_cell:
+          - background-color: "[[[ return `rgba(var(--color-${variables.color}), 0.2)`; ]]]"
+        card:
+          - background-color: "[[[ return `rgba(var(--color-background-${variables.color}), var(--opacity-bg))`; ]]]"
+  tap_action:
+    action: "call-service"
+    service: "input_select.select_option"
+    service_data:
+      entity_id: "[[[ return entity.entity_id ]]]"
+      option: "[[[ return variables.state ]]]"
 
 ### CARDS ###
 card_generic:
--- a/config/minimalist-templates/button_card_templates.yaml
+++ b/config/minimalist-templates/button_card_templates.yaml
@@ -276,9 +276,20 @@ card_title:
       - font-size: "1rem"
       - opacity: "0.4"
 ### VERTICAL BUTTONS (fka SCENES) ###
-vertical_buttons:
+card_vertical_button:
+  variables:
+    ulm_card_vertical_button_color: "blue"
+    ulm_card_vertical_button_state: "on"
   show_label: true
-  label: "[[[ return (entity.attributes.value )]]]"
+  label: ""
+  name: |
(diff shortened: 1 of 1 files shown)


mp-se:
It looks like my problem is connected to this, https://github.com/custom-cards/button-card/issues/506

mp-se:
There are apparently limitation what you can nest in templates. I had to put the code that handles the old variable names in the base template. not the way i would like to have it. but it does not work otherwise. i put a comment in there so when the old templates are removed we can remove that code as well.

config/minimalist-templates/button_card_templates.yaml
```diff
@@ -275,10 +275,34 @@ card_title:
...
+          if(typeof variables.color !== 'undefined')
+            return variables.color;
+         return "blue";
+      ]]]
+    ulm_card_vertical_button_state: |
+      [[[
+        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
+          if(typeof variables.color !== 'undefined')
+            return variables.state;
+         return "on";
```
saxel, line 294: Indentation error
```suggestion
        return "on";
```

config/minimalist-templates/button_card_templates.yaml
```diff
@@ -275,10 +275,34 @@ card_title:
...
+      [[[
+        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
+          if(typeof variables.color !== 'undefined')
+            return variables.color;
+         return "blue";
+      ]]]
+    ulm_card_vertical_button_state: |
+      [[[
+        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
+          if(typeof variables.color !== 'undefined')
```
saxel, line 292:
```suggestion
          if(typeof variables.state !== 'undefined')
```

config/minimalist-templates/button_card_templates.yaml
```diff
@@ -275,10 +275,34 @@ card_title:
...
 ### VERTICAL BUTTONS (fka SCENES) ###
-vertical_buttons:
+card_vertical_button:
+  variables:
+    ulm_card_vertical_button_color: |
+      [[[
+        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
+          if(typeof variables.color !== 'undefined')
+            return variables.color;
+         return "blue";
```
saxel, line 287:
```suggestion
        return "blue";
```

saxel (commented):
I tried your changes and it doesn't seem like the compatability mode works still?

image

If it is not possible to solve then I think there is no reason to keep the custom_state template.

mp-se:
Let me check, if I missed something copying the changes from my HA directory.

mp-se:
i've tested the code on my HA now and for me it works. Perhaps I have used the orginal cards diffrently, this is my test configs:
```
- type: 'custom:button-card'
  template: card_vertical_button
  name: Kitchen
  entity: input_boolean.scene_kitchen_work

- type: 'custom:button-card'
  template: card_vertical_button
  entity: input_select.test_vertical_buttons
  name: Away
  icon: mdi:television-classic
  variables:
    ulm_card_vertical_button_state: Away

- type: 'custom:button-card'
  template: card_vertical_button
  entity: input_select.test_vertical_buttons
  name: Night
  icon: mdi:television-classic
  variables:
    ulm_card_vertical_button_state: Night
    ulm_card_vertical_button_color: green

- type: 'custom:button-card'
  template: vertical_buttons
  name: Kitchen (legacy)
  entity:  input_boolean.scene_kitchen_work

- type: 'custom:button-card'
  template: vertical_buttons_custom_state
  entity: input_select.test_vertical_buttons
  name: Night2 (legacy)
  variables:
    state: "Night"
    color: "red"

- type: 'custom:button-card'
  template: _vertical_buttons_custom_state
  entity: input_select.test_vertical_buttons
  name: Night2 (legacy olg)
  variables:
    state: "Night"
    color: "red"

- type: 'custom:button-card'
  template: vertical_buttons_custom_state
  entity: input_select.test_vertical_buttons
  name: Night2 (legacy)
  variables:
    state: "Night"

- type: 'custom:button-card'
  template: _vertical_buttons_custom_state
  entity: input_select.test_vertical_buttons
  name: Night2 (legacy olg)
  variables:
    state: "Night"
```

saxel:
Ah, yeah I used the variable.state to display as the name as well. But that's fine. Does activating the legacy button work for you?

mp-se:
> Ah, yeah I used the variable.state to display as the name as well. But that's fine. Does activating the legacy button work for you?

No, it doesent do anything, i guess that was a thing paddy was going to adress.

CM000n:
what do you guys think, is this ready to merge, or rather in a WIP status?

mp-se:
> what do you guys think, is this ready to merge, or rather in a WIP status?

According to how i interpret the design, it works as before so it can be merged.  I have not found any docs but if anyone has other views i can test/update according to that input.

saxel:
Yeah the updated card works, but not the compatibility ones. All it results in is the "dummy" buttons that doesn't work, no color on active state, and no actions happen when pressed.

You have code for compatibility mode in the updated card, but it doesn't work. the variables are always undefined. Is it supposed to work?
```yaml
    ulm_card_vertical_button_color: |
      [[[
        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
          if(typeof variables.color !== 'undefined')
            return variables.color;
         return "blue";
      ]]]
    ulm_card_vertical_button_state: |
      [[[
        if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
          if(typeof variables.color !== 'undefined')
            return variables.state;
         return "on";
      ]]]
```

mp-se:
> Yeah the updated card works, but not the compatibility ones. All it results in is the "dummy" buttons that doesn't work, no color on active state, and no actions happen when pressed.
> 
> You have code for compatibility mode in the updated card, but it doesn't work. the variables are always undefined. Is it supposed to work?
> 
> ```yaml
>     ulm_card_vertical_button_color: |
>       [[[
>         if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
>           if(typeof variables.color !== 'undefined')
>             return variables.color;
>          return "blue";
>       ]]]
>     ulm_card_vertical_button_state: |
>       [[[
>         if(typeof variables !== 'undefined' ) // This code is for backwards compatibility
>           if(typeof variables.color !== 'undefined')
>             return variables.state;
>          return "on";
>       ]]]
> ```

It works on my installation, so we might be using them diffrently. Can you share how you use the cards so I can test those scenarios ?

saxel:
I tested with your examples that were using input_select, but switched out the entity and the state value to my own, also tested with name attribute removed. Something weird happens when the page is first loaded. First nothing happens when I click the buttons. Then I click a button which has been has been migrated to card_vertical_button. Suddenly the label of the legacy button changes from "on" to the state value. And now the first two legacy buttons work, but not the last two without color variable. I'm not sure what is going on here.

vertical_buttons_custom_state3

```yaml
              - type: 'custom:button-card'
                template: vertical_buttons_custom_state
                entity: input_select.status_saxel
                name: Night2 (legacy)
                variables:
                  state: "home"
                  color: "red"

              - type: 'custom:button-card'
                template: vertical_buttons_custom_state
                entity: input_select.status_saxel
                variables:
                  state: "away"
                  color: "red"

              - type: 'custom:button-card'
                template: vertical_buttons_custom_state
                entity: input_select.status_saxel
                #name: Night2 (legacy)
                variables:
                  state: "exercise"

              - type: 'custom:button-card'
                template: vertical_buttons_custom_state
                entity: input_select.status_saxel
                name: Night2 (legacy olg)
                variables:
                  state: "work"
```

mp-se:
Wonder if this is connected to that I use service calls to switch states , since it needs to be processed via HA event handling this could cause a delay before the state change is detected. On my system the delay is not visible. What are you running you HA instance on? But if there are issues it might be better to just add the new card and deprecate the other vertical buttons?

saxel:
I don't think there's an issue with delays since the legacy buttons starts working after I press any migrated button. My HA is running in a VM with plenty of resources available so I don't think that is an issue either.

Every time I reload/refresh the page I can reproduce it like this

mp-se:
This is how it looks on my system

vb

saxel:
Super weird, I don't know why it would work for you but not for me 🤷‍♂️I would anyway just migrate straight away, but I was curious about the legacy/compatibility card.

mp-se:
> Super weird, I don't know why it would work for you but not for me 🤷‍♂️I would anyway just migrate straight away, but I was curious about the legacy/compatibility card.

I agree, its not a good sign if it behaves differently that always makes me uncomfortable.  Perhaps I should keep the legacy templates for now and just introduce the new card ?

mp-se:
To support issue #198 there are code changes needed. It currently only supports input_boolean and input_select as data types. Its not difficult to add those but I would like to have a decision on what is the next step regarding backwards compatibility

saxel:
If we absolutely don't want to break compatibility then we need to keep the legacy template for at least 1 release.

CM000n:
Since the next release will probably contain a lot of breaking changes anyway, I think we can make a cut here and go the new way and deprecate the old templates.
If there are issues popping up regarding incompatibility and no longer working vertical buttons, we can then refer them to the new templates.

mp-se:
I will fix this later today then, i will probably align this with the new structure as well. so i might skip this PR and open a new one.

mp-se:
I've done the updates and also added so entities of type light & switch can be handled. That should support #198 if the user adds show_name: false to their defintion

PR #230, opened 2022-01-08
HACS: more rename Fixes

stokkie90:


.gitignore
```diff
@@ -5,6 +5,7 @@ ui_lovelace_minimalist/
 /themes/
 
 !custom_components/ui_lovelace_minimalist
+!custom_components/__init__.py
```
CM000n, line 8: Why are you ignoring the ini file now @stokkie90?

stokkie90: It's not ignoring it! But there are some problems with the latest release of hacs itself. So I started out thinking it was caused by some changes I made.  Bit that was not the case, but I did find some.minor rename things I missed.

CM000n: Ahh, sorry, I misread it

PR #233, opened 2022-01-08
Custom Bar Card

rphlwnk:
I created a minimalist template for the `custom-card` [bar-card](https://github.com/custom-cards/bar-card) to display a (percentage) values.

I sticked mostly to the last release custom cards regarding the stucture, please let me know if it does not fit.

screenshot


CM000n:
Hi, I think it looks great. :-) Thank you very much! 
To be honest, I just don't know if the appearance of the bar matches the design of the rest of the theme. Otherwise, this would also be a good card for the standard templates.
But as a custom card it's definitely fine!

rphlwnk:
Thanks! :-)

In my setup i think it blends nicely with the graph_card, since both are using the "no edge" style:

image

CM000n:
Cool :-) I will definitely add them to the custom cards, maybe later we can think about adding them to the standard templates if it's ok with you.
With HACS it doesn't make a big difference anymore anyway, because then all cards, no matter if standard or custom, will be downloaded and updated anyway :-D

PR #231, opened 2022-01-08
Button Cards Templates split

sildehoop:
I have updated my pr #67
Beginning to move to new file structure https://github.com/UI-Lovelace-Minimalist/UI/discussions/105#discussioncomment-1866879

PR #242, opened 2022-01-10
Further folder and template file restructuring for popup cards

CM000n:
Following https://github.com/UI-Lovelace-Minimalist/UI/pull/231, we should probably split the config files for our popup templates as well


schumijo (approved):
Perfect work ! Everything works fine

PR #245, opened 2022-01-11
Documentation

schumijo:
Quick and dirty copy paste from paddy documentation
Need some review on formatting
To be updated with new install method, new cards, etc..
But good be a good basis

Just enable Github pages on repo


CM000n:
I close this one, because we are now trying to make our own documentation with mkdocs https://github.com/UI-Lovelace-Minimalist/UI/pull/250

PR #252, opened 2022-01-12
Documentation improvements

stokkie90:


.github/workflows/pages.yml
```diff
@@ -2,7 +2,7 @@
 name: "Pages"
 
 on:
-  pull_request:
+  push:
```
CM000n, line 5: I changed that on purpose. Most changes here happen via pull_requests and not pushes. Especially not in the "main" branch.

docs/usage/cards/._example.md
```diff
@@ -0,0 +1,25 @@
...
+## __card_name_1__
+
+Card description should be here.
+
+|Variable|Default|Required|Notes|
+|----------|----------|------|------|
+|name| | X | |
+|background| `true` | X | Enable background |
+
+![example-image](img/example.png)
```
CM000n, line 14: I don't know if I like the idea of each subdirectory containing its own assets. I'm more in favour of putting all the assets in a central folder in the root directory of the docs. Any other opinions on this?

.github/workflows/pages.yml
stokkie90: This won't work. As it only needs to build pages on those branches and not every pull request pointing to those branches. See GitHub actions before last commit.

docs/usage/cards/._example.md
stokkie90: It can also be put under assets. But could make it harder to get the correct file. From my experience this works pretty well.

CM000n: If you look at cms like wordpress or drupal, etc., they don't have an asset hanging on every page but have a central place for assets that are referenced from the respective page.
There are maybe situations where the same asset is used on multiple pages. Then you would have to either copy it unnecessarily or create strange cross-references.

stokkie90: That is true indeed. So should we go for assets/img/..

.github/workflows/pages.yml
CM000n: Got it now. Thx for your explanation!

PR #254, opened 2022-01-13
DRAFT: Add support for awesome-pages-plugin

stokkie90:


mkdocs.yml
```diff
@@ -46,6 +38,7 @@ theme:
     - "navigation.instant"
     - "navigation.tracking"
     - "navigation.top"
+    - "navigation.indexes"
```
CM000n, line 41: Regarding the material documentation this feature flag doesn't work in combination with `toc.integrate`
https://squidfunk.github.io/mkdocs-material/setup/setting-up-navigation/#integrated-table-of-contents

mkdocs.yml
```diff
@@ -8,23 +8,15 @@ repo_url: "https://github.com/UI-Lovelace-Minimalist/UI"
...
-      - "usage/index.md"
-      - Cards:
-          - "usage/cards/index.md"
-          - Light: "usage/cards/light.md"
-  - Development:
-      - "development/index.md"
-      - Docs Style Guide: "development/docs_style_guide.md"
+  - "... | setup/**/*.md"
+  - "... | usage/**/*.md"
+  - "... | development/**/*.md"
```
CM000n, line 16: This may look tidier here. But now I have to maintain each order in the respective folder separately, if I see it right and have no central place for it anymore?

mkdocs.yml
stokkie90: Yeah but the strange thing is, it works :laughing: at least local testing. If you check out this branch locally you can see test it.

CM000n: Hehe cool, so let‘s try it 😉

PR #257, opened 2022-01-14
Add example wiki files for every card chip and popup

CM000n:


CM000n:
Get's maybe obsolete because of https://github.com/UI-Lovelace-Minimalist/UI/pull/257

PR #258, opened 2022-01-14
Add images

sildehoop:
Following #242, move further to new file structure #105 (reply in thread)
And added the images for the Wiki


CM000n:
Hi @sildehoop , really great work. Thank you very much!
I have seen that for each card you have placed the corresponding images and *.md files in a respective folder of the card within the custom_component structure.
In order to keep the data consumption as low as possible during the download and installation via HACS, we wanted to store the assets and dokus outside the custom_component in the docs directory.

You should therefore place all images directly here (you are also welcome to delete superfluous ones): https://github.com/UI-Lovelace-Minimalist/UI/tree/dev/docs/assets/img
All *md files for the documentation of the cards should be stored in the respective directories here:https://github.com/UI-Lovelace-Minimalist/UI/tree/dev/docs/usage

PR #261, opened 2022-01-15
Add images

sildehoop:
Added the images form the Figma


CM000n:
wonderful! Thank you very much sil

PR #262, opened 2022-01-16
Update NL.yaml

basbruss:
Correct dutch translations
Do have we a design-standard to use uppercase or lowercase? It's now a mix of both.


CM000n:
Thanks @basbruss! 
Currently, we do not have a upper- / lowercase design-standard yet. 
Whereas I would definitely be in favor of starting with capital letters. But maybe that's just my German biased view of things 😉

PR #267, opened 2022-01-18
Add github workflow for automatic custom_cards docs

CM000n:


sildehoop:
Why don't we do this with the normal cards too??

CM000n:
> Why don't we do this with the normal cards too??

Because the "standard" template .yaml files are delivered with the custom_component and we want to keep the download size for HACS as low as possible.
In this case we have to save templates and docs separately anyway and can store the *.md files directly in the docs.

PR #274, opened 2022-01-19
Initial start on dashboarding

stokkie90:


rick2909:
just a tip if you dont want the pull request not be merged. Use the Draft feature from github
[Converting a pull request to a draft](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft)

CM000n:
> @rick2909 Thanks! used to gitlab 😅

It‘s the same in gitlab 😉

.pre-commit-config.yaml
```diff
@@ -20,6 +20,11 @@ repos:
       - id: "check-merge-conflict"
       - id: "check-yaml"
         args: ["--unsafe"]
+        exclude: |
+            (?x)^(
+                custom_components/ui_lovelace_minimalist/installation/.*\.yaml|
```
CM000n, line 25: Why do you add these folders/data to be ignored by "chack-yaml"?

stokkie90:
Not entirely though. Gitlab works if you add WIP or DRAFT  in the title. 

But indeed not possible to merge works the same. Just didn't know GitHub had this feature.

.yamllint
```diff
@@ -32,7 +32,7 @@ rules:
   new-lines: "enable"
   octal-values: "disable"
   quoted-strings:
-    quote-type: "double"
+    quote-type: "any"
```
CM000n, line 35: I am definitely against this adjustment. Our .yaml files follow the Home Assistant Style Guide for .yaml files, which prescribes double-quoted strings. I see no reason why we should not apply this to the rest of the .yaml files in the repo as well

.pre-commit-config.yaml
```diff
@@ -33,7 +38,13 @@ repos:
     hooks:
       - id: "yamllint"
         args: ["--format", "parsable", "--strict"]
-        exclude: ".pre-commit-config.yaml"
+        # exclude: ".pre-commit-config.yaml"
+        exclude: |
+            (?x)^(
+                .pre-commit-config.yaml|
+                custom_components/ui_lovelace_minimalist/installation/.*\.yaml|
```
CM000n, line 45: Why do you add these folders/data to be ignored by "yamllint"?

.pre-commit-config.yaml
stokkie90: Because of Jinja templating and this check not liking it. 😅

.yamllint
stokkie90: Check, will reverse it.

.pre-commit-config.yaml
stokkie90: Jinja templating.

.pre-commit-config.yaml
stokkie90: Might think of changing it to extension based. But requires some more work.

CM000n: Ahh, good to know.
What's exactly the problem with jinja templating?
What error is it throwing?

stokkie90:
```
while scanning for the next token
found character that cannot start any token
  in "custom_components/ui_lovelace_minimalist/lovelace/views/main/rooms/room.yaml", line 5, column 2
while scanning for the next token
found character that cannot start any token
  in "custom_components/ui_lovelace_minimalist/lovelace/views/main/01_home.yaml", line 8, column 6
````

PR #277, opened 2022-01-20
Bug fixes card_light and icon_more_info

basbruss:


AndyVRD:
Just tested the new all in one light_card with the updated icon_more_info and wen an entity is unavailable the slider is showing in grey? The notification icon is working fine now btw.

PR #278, opened 2022-01-20
added dark mode box shadow custom_card_scenes

sildehoop:
added the option for dark mode box shadow
Schermafbeelding 2022-01-20 om 13 24 19


CM000n:
What happens if this new option is set to true and I am in the light theme?
For example, because I have an automatic switch between light and dark theme based on the time.

AndyVRD:
Tested but if i enable dark_mode: true the dark box-shadow is always showing, also if my light theme is active and the sun.sun is above_horizon. I think there is still something wrong? Thx btw for adding this option. ;)

saxel:
What if the user chose permanent Dark or Light mode? It would be weird to get different shadow depending on the Sun 😀

Can't we use some variable from the theme to determine if it is Dark mode or Light mode?

sildehoop:
I didn't think about the possibility of having automatically selected theme so I only have the light theme selected.

https://github.com/custom-cards/button-card#javascript-templates

Inside the javascript code, you'll have access to those variables:

`this:` The button-card element itself (advanced stuff, don't mess with it)
`entity:` The current entity object, if the entity is defined in the card
`states:` An object with all the states of all the entities (equivalent to hass.states)
`user:` The user object (equivalent to hass.user)
`hass:` The complete hass object
`variables:` an object containing all your variables defined in the configuration. See Variables

I don't know how to use the `user:` and `hass:` variables and what information they give you but Maybe we can get the current select it theme from there.

basbruss:
In the card_light we made a solution with support for light, dark and auto mode(depending on sun elevation). Maybe you can use the same approach in this card to change the box-shadow.

saxel:
You can use `hass.themes.darkMode` in the if-statement. It returns true if dark mode is enabled. It seems to work with the Auto setting as well.

saxel:
Yeah I also had problems finding it. There is some mention about it on the home assistant github

https://github.com/home-assistant/frontend/blob/303e065433b720f0de256069beda7bd87dd33e96/src/types.ts#L89

```javascript
// Currently selected theme and its settings. These are the values stored in local storage.
// Note: These values are not meant to be used at runtime to check whether dark mode is active
// or which theme name to use, as this interface represents the config data for the theme picker.
// The actually active dark mode and theme name can be read from hass.themes.
export interface ThemeSettings {
  theme: string;
  // Radio box selection for theme picker. Do not use in Lovelace rendering as
  // it can be undefined == auto.
  // Property hass.themes.darkMode carries effective current mode.
  dark?: boolean;
  primaryColor?: string;
  accentColor?: string;
}
```
As I read it using hass.themes is OK

basbruss:
I've already tried it with the colored light card and it seems te work as intended 😄 
```
if (hass.themes.darkMode){
  if (color){
    return 'rgba(' + color + ',0.1)'
  } else {
    return 'rgba(var(--color-yellow),0.1)'
else {
  return 'rgba(var(--color-background-yellow),var(--opacity-bg))';
}
```

AndyVRD:
Thanks @basbruss that works perfect! Didn't even know this was possible and is way better than using the sun.sun state i think. 

```
  - box-shadow: >
      [[[
        if (hass.themes.darkMode){
          return "0px 2px 4px 0px rgba(0,0,0,0.80)";
        } else {
          return "var(--box-shadow)";
        }
      ]]]```

PR #279, opened 2022-01-20
Add option to enable/disable side panel entry during Integration installation

stokkie90:
Added option to enable/disable Sidepanel entry for the generated dasbhoard, Default it will be on False.

Also added Mobile detection, if mobile move nav-bar to the bottom.


.pre-commit-config.yaml
```diff
@@ -33,7 +38,13 @@ repos:
     hooks:
       - id: "yamllint"
         args: ["--format", "parsable", "--strict"]
-        exclude: ".pre-commit-config.yaml"
+        # exclude: ".pre-commit-config.yaml"
+        exclude: |
+            (?x)^(
+                .pre-commit-config.yaml|
+                custom_components/ui_lovelace_minimalist/installation/.*\.yaml|
```
CM000n, line 45: Maybe this is a better way to ignore only certain sections explicitly, instead of ignoring complete files or folders:
https://yamllint.readthedocs.io/en/stable/disable_with_comments.html?highlight=jinja#disabling-checks-for-all-or-part-of-the-file

PR #282, opened 2022-01-21
Enable popup on new AIO light card

schumijo:
With the new AIO light card, light popup was not working. 
Added variable to enable/disable popup.
Documentation updated


basbruss:
Nice adjustments!
Is it not better to add the popup to the `hold_action` as all popups do? And also add it to the text-area and the main card instead of only the icon. So they can hold everywhere to get the popup.
Before it did also work, but only around the whitespace-edges of the card.

schumijo:
See #283

PR #286, opened 2022-01-22
Create custom card input datetime

sildehoop:
Schermafbeelding 2022-01-22 om 17 28 51

### DIFF
--- a/custom_cards/custom_card_input_datetime/README.md
+++ b/custom_cards/custom_card_input_datetime/README.md
@@ -1,0 +1,62 @@
+---
+title: card_input_datetime
+---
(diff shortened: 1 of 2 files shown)


CM000n:
Cool, once again a great job @sildehoop. Whereas I would probably show the middle "button" (widget_text) without background, if it is not a button. As madmax once noted in Discord, you should only suggest interaction possibilities where they are realy possible.

sildehoop:
It has an press and hold action press is +1 min and hold -1 min

CM000n:
> It has an press and hold action press is +1 min and hold -1 min

Ahh, very cool! thx for clarification.

PR #289, opened 2022-01-23
Improve media player render with diverse entity_picture

mazerte:
Hello 👋 ! 

Thanks for your awesome work. This PR is to improve the media player with more diverse entity_picture, like very dark or very light or busy images.

Question: will `card_media_player_with_control` is call like that? Because it doesn't propose any feature to display controls. I thing removing `card_media_player` and replace it with this one could be less confusing. For backward compatibility we can add a variable to enable picture background.

Let me know what you think and a will add it to the PR.


CM000n:
Hi @mazerte, thank you very much for your MR. Yes, the current naming of the media_player is a bit confusing. Our future aspiration should be to have only one media_player card, and simply customize its representation via variables. Similar to what @basbruss and @schumijo have done for the card_light thanks to their great work.
If you would like to take care of this, we would of course be more than happy. Then I would put your MR into draft status for the time being and you could continue working on it.

mazerte:
I take it because I want to create one for myself 😄 .
I will also include Sonos grouping options in.

CM000n:
Cool! If you need any help, or would like some advice, feel free to drop by our Discord channel and visit us in the #development area

schumijo:
Hi @mazerte 

As @CM000n said, if you need help you can take some inspiration from our light card or also ask me🇨🇵😉. 

I was also thinking about a new player card that merge some custom ones : Chromecast, so is, playstation, etc...But I am actually focus on new all in one thermostat card with popup (teaser 🤣)

mazerte:
card_media_player_art_controls
card_media_player_art
card_media_player_controls
card_media_player_icon
card_media_player

mazerte:
I also manage the backward compatibility so it shouldn't be an issue

basbruss (approved):

PR #290, opened 2022-01-23
Fix simple weather card text alignment

mazerte:
Hello 👋 !

I notice that the text in the `simple-weather-card` wasn't well aligned. So I fix it.

Thanks


mazerte:
I just add more options to use backdrop and customize the display name. Doc updated.

PR #295, opened 2022-01-23
Add chip images

sildehoop:
I have added the image of the chips mentioned in #29
chip_alarm
chip_back
chip_icon_double_state
chip_icon_label
chip_icon_only
chip_icon_state
chip_mdi_icon_only
chip_mdi_icon_state
chip_navigate
chip_power_consumption
chip_presence_detection
chip_temperature


CM000n:
cool, thanks for the work @sildehoop. I think in this case we can merge it directly into "Main" so that the changes are directly in the doc.

Issue #294, opened 2022-01-23
Add missing images to documentation


aehru:
Hello,

**Is your enhancement request related to a problem? Please describe.**
I'm reviewing the documentation (chips part right now) but some images are missing on the docu and in the Figma

**Describe the solution you'd like**
Add image to Figma for those chips :
- [x] chip_alarm
- [x] chip_back
- [x] chip_icon_dauble_state (exists but currently named chip_icon_only)
- [ ] chip_icon_label (typo Lable)
- [x] chip_icon_only 
- [x] chip_icon_state
- [x] chip_navigate
- [x] chip_power_consumption
- [x] chip_presence_detection
- [x] chip_temperature

PR #301, opened 2022-01-23
fix typo lable to label

sildehoop:
thanks to #294


sildehoop:
Saw it too late to add it to the other pr

PR #311, opened 2022-01-24
Add primary and secondary info vars to card_weather

willyp713:
Adds two new variables for `primary_info` and `secondary_info` for the `custom:simple-weather-card` that is the foundation for this card. More details about these fields is available on the [Simple Weather Card repository](https://github.com/kalkih/simple-weather-card).

This will also need a corresponding change in the `card_weather` documentation to show that these variables are available and how to use them.


CM000n:
Nice! Can you also adjust the documentation of the card according to your changes?

willyp713:
Actually found a small issue with this. The "default" behavior doesn't seem to work if you choose not to pass through a value for the variable from your UI configuration.

If you _do_ specify a value for either of these variables, then they will show that data correctly.

willyp713:
Ok new minor change; provided initial default values for the two new variables at the top of the file. That appears to fix the default behavior bug.

willyp713:
And lastly merged in the docs updates into this same PR.

CM000n:
Nice! Thank you very much for you work!

PR #319, opened 2022-01-25
Started on initial dashboard creation and jinja2 support

stokkie90:


.yamllint
```diff
@@ -17,8 +17,7 @@ rules:
   comments-indentation:
     level: "warning"
   document-end: "disable"
-  document-start:
-    present: true
+  document-start: "disable"
```
CM000n, line 20: why?
https://yaml.org/spec/1.1/#id897596

stokkie90: `# ui_lovelace_minimalist` must be on the first line and can't be excluded with another comment. And it's a nice to have in yaml.

CM000n: :-/

Issue #318, opened 2022-01-25
Card_light pop up fails; falls back to more_info


willyp713:
**Describe the bug**
Card light pop up window fails to open. The long press/hold action appears to do nothing. The card shows a gray highlight and indicates it is being pressed, but does not open the pop up. Short press or tap opens `more-info`.

**To Reproduce**
Extra install requirements:
1. Install [hass-browser_mod](https://github.com/thomasloven/hass-browser_mod) in HACS
2. Install [my-cards](https://github.com/AnthonMS/my-cards) in HACS
2. Install [light-entity-card](https://github.com/ljmerza/light-entity-card) in HACS

Create card in UI config:
1. Add `card_light` card and configure for basic operation
2. Include variable `ulm_card_light_enable_popup: true`
3. Refresh browser window and test `hold_action` on new card
4. **Problem:** card shows it is being pressed (gray highlight over icon or text) but nothing happens
5. Alternatively, tapping activates `more-info`

**Expected behavior**
Pop-up window appears on `hold_action` with `light-entity-card` controls

**Screenshots**

https://user-images.githubusercontent.com/21319282/150919708-04a8e9b1-a9fd-4b66-b2c6-4e5248411529.mp4

**Additional context**
Tested in iOS Home Assistant App and Brave browser on Windows and iOS

Sample card code:
```
- type: "custom:button-card"
  template: 
    - card_light
    - popup_light
  entity: light.kitchen_island
  variables:
    ulm_card_light_enable_slider: true
    ulm_card_light_enable_collapse: true
    ulm_card_light_enable_popup: true
```

basbruss:
Could you try the config without adding the popup_light template:
```
- type: "custom:button-card"
  template: 
    - card_light
  entity: light.kitchen_island
  variables:
    ulm_card_light_enable_slider: true
    ulm_card_light_enable_collapse: true
    ulm_card_light_enable_popup: true
```
As the card is developed to only set the variable to use the popup, where adding a template can overrule some functions.

CM000n:
Maybe stupid question but did you add `browser_mod:` to your `configuration.yaml` like described in the docs there?

willyp713:
> Could you try the config without adding the popup_light template:
> 
> ```
> - type: "custom:button-card"
>   template: 
>     - card_light
>   entity: light.kitchen_island
>   variables:
>     ulm_card_light_enable_slider: true
>     ulm_card_light_enable_collapse: true
[...]

PR #324, opened 2022-01-25
Clearify usage

basbruss:
Add warning and add suggestion from #318


basbruss:
Changed exemple to example in all popups.md

PR #325, opened 2022-01-25
added missing english translations

rick2909:
noticed 2 custom cords missed the EN language files and the welcome card the ES language file was english. this is now correct spannish and moved the english translation to EN.yaml file


CM000n:
Thank you very much! :-)

PR #338, opened 2022-01-26
Add welcome_card for auto-generated dashboard

basbruss:
Add `card_welcome_scenes`
Add translations to main files
image


rick2909:
it looks liike it is the same one that is now a custom one, maybe remove the custom card.

basbruss:
@stokkie90 could you test these changes in your dev-env? My system had some hiccups with loading of javascript and the templates, but the code should be correct.

AndyVRD:
Thanks for the merge and release!

basbruss:
Formated date chip to:
image

stokkie90:
@basbruss i tested it on my dev instance:

- After installation, first time you will get, can't find the cards.  (caching etc) but after ctrl + f5 and/or refresh (top-right-corner) it load the dashboard.
- After installation no restart is required. but it will not load themes. so we might need to extend docs a bit. with the action to restart or reload themes. using the commands  (hit "c" on keyboard)

custom_components/ui_lovelace_minimalist/lovelace/ui-lovelace.yaml
```diff
@@ -29,10 +29,14 @@ views:
...
                   options:
                     type: "custom:button-card"
-                    template: "card_paddy_welcome_with_weather"
+                    template: "card_welcome_scenes"
                     variables:
-                      ulm_card_paddy_welcome_time: "sensor.time"
-                      ulm_card_paddy_welcome_weather_provider: "this.entity_id"
+                      ulm_time: "sensor.time"
+                      ulm_weather: "this.entity_id"
+          - type: "custom:button-card"
```
stokkie90, line 36: Double title "Lights"

stokkie90 (approved):

PR #339, opened 2022-01-27
card_light: Use icon if it's set on the device

stokkie90:
If not specified should be using `mdi:lightbulb`. 

Needs testing. or we should move it to a if/else statment


r0binj:
It works great, I think its better to get the icon from the entity instead of have it in the code... Less code is better ;)
Skärmavbild 2022-01-27 kl  09 40 42

PR #86, opened 2021-12-15
Fix inconsistent naming with_

sildehoop:
With the card_cover_with_buttons is **with** used and card_light_slider not

Template name: card_battery
Template name: card_binary_sensor
Template name: card_binary_sensor_alert
Template name: card_cover_with_buttons **--> card_cover_buttons**
Template name: card_generic
Template name: card_generic_swap
Template name: card_graph
Template name: card_input_boolean
Template name: card_light
Template name: card_light_slider
Template name: card_light_slider_collapse
Template name: card_light_slider_horizontal
Template name: card_media_player
Template name: card_media_player_with_control **--> card_media_player_art**
Template name: card_media_player_with_controls **--> card_media_player_controls**
Template name: card_navigate
Template name: card_person
Template name: card_power_outlet
Template name: card_script
Template name: card_title
Template name: chip_back
Template name: chip_icon_double_state
Template name: chip_icon_only
Template name: chip_icon_state
Template name: chip_power_consumption
Template name: chip_presence_detection
Template name: chip_temperature

PR #342, opened 2022-01-27
Fix card naming

sildehoop:
Remake of #86
I have recently seen more confusion around the card_media_player_controls and card_media_player_control naming. I think it is beter to adres this now instead of delaying it.

Template name: card_battery
Template name: card_binary_sensor
Template name: card_binary_sensor_alert
Template name: card_cover_with_buttons --> card_cover_buttons
Template name: card_generic
Template name: card_generic_swap
Template name: card_graph
Template name: card_input_boolean
Template name: card_light
Template name: card_light_slider
Template name: card_light_slider_collapse
Template name: card_light_slider_horizontal
Template name: card_media_player
Template name: card_media_player_with_control --> card_media_player_art
Template name: card_media_player_with_controls --> card_media_player_controls
Template name: card_navigate
Template name: card_person
Template name: card_power_outlet
Template name: card_script
Template name: card_title
Template name: chip_back
Template name: chip_icon_double_state
Template name: chip_icon_only
Template name: chip_icon_state
Template name: chip_power_consumption
Template name: chip_presence_detection
Template name: chip_temperature


CM000n:
Seems you mixed up your changes on you inpu number card together with the renamings?
Makes the PR harder to read.

But yeah, I'm absolutely with you.
People are confused and we have to change it better sooner then later

sildehoop:
Ohh sorry didn't notice

sildehoop:
update #349 after push or vice versa

CM000n (approved):

CM000n:
If I see it correctly, this will break previous uses of the cards, right?
Is there a way to maintain compatibility with the old naming convention or do we just want to make a hard cut and put some information about it in the release notes?

sildehoop:
Yes your right and I don't now it it is the right decision to keep dragging the old templates and try to support them. if jou want to improve you need to have braking changes. I will try making a legacy templates with the old naming and add the renaming for #349

CM000n:
@stokkie90 how do you feel about it?
In my opinion we should go for a hard cut here

CM000n:
Think we can merge it @sildehoop 
E
We should just take care of a proper breaking change message in the release noted

CM000n (approved):

sildehoop:
should I remove the legacy templates ??

CM000n:
> should I remove the legacy templates ??

Any more opinions on that @schumijo, @bavo ..?

schumijo:
> > should I remove the legacy templates ??
> 
> Any more opinions on that @schumijo, @bavo ..?

Maybe we could keep the legacy template for now and then removed them in 2 or 3 release as done by HA with deprecated integration.
It's just an idea...

bavo:
Yeah maybe keep them, but deprecate them in the release notes

CM000n:
I've merged it now and noted it in the Release Note Draf here: https://github.com/UI-Lovelace-Minimalist/UI/pull/337
Please add/correct if needed.

PR #349, opened 2022-01-28
Correcting the titles in the doc(s)

r0binj:
Correcting the title in template code on the docs.


sildehoop (approved):

sildehoop:
update #342 after push or vice versa

r0binj:
Need to change they title in the cards that you're renaming so it matches the new 😋

Looked at the custom cards and saw that the most of them missing the template on wiki. I can go through them later this evening.

PR #359, opened 2022-01-30
Fix translations

aehru:
Complete FR translation + fix displayed state of card_person. card_person uses variables["ulm_person_state_" + state] and Home Assistant state for "away" seems to be "not_home".
I don't know if it's required to keep ulm_person_state_away entry. Does someone configuration uses it ?


stokkie90 (approved):
Lets leave the state_away for now.

PR #368, opened 2022-01-31
Fix card_battery charging icon

AndyVRD:
For me the mdi:battery-charging icon was not working for iOS/iPad OS. The state of there battery_state sensor is "Charging" if connected to a power outlet. But a close look at the code and far as i understand it have a problem i think. If the variables.ulm_card_battery_charger_type_entity_id is not defined by the user or the state is not "null", "ac" or "wireless" the battery-charging icon will never be used. 

The strange thing is that the code below (far as i understand...) will pass the battery-charging icon if the state of variables.ulm_card_battery_battery_state_entity_id is not equal to "null" and the state of variables.ulm_card_battery_battery_state_entity_id is "charging" but for some reason this was not working for me (also after changing "charging" to "Charging"...).

 infix = variables.ulm_card_battery_battery_state_entity_id !== null
        && states[variables.ulm_card_battery_battery_state_entity_id].state == "charging"
        ? "-charging" : ""


custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_battery.yaml
```diff
@@ -28,14 +28,17 @@ card_battery:
       if(variables.ulm_card_battery_charger_type_entity_id == null) {
         // is charging?
         infix = variables.ulm_card_battery_battery_state_entity_id !== null
-                && states[variables.ulm_card_battery_battery_state_entity_id].state =="charging"
+                && states[variables.ulm_card_battery_battery_state_entity_id].state == "charging"
```
checkerschaf, line 31: Could you use `===` for string equality checks? 😄 I think we should update the code base in general to always use `===` where applicable.

CM000n: Bit more background would be nice :-)
https://codeahoy.com/javascript/2019/10/12/==-vs-===-in-javascript/#:~:text=The%20difference%20between%20%3D%3D%20and%20%3D%3D%3D%20is%20that%3A,the%20two%20variables%20being%20compared.

bavo: We could also covert the state to lowercase and then compare to "charging". This should also remove all problems related to "Charging/charging"

CM000n: Hi @AndyVRD. Do you want to integrate the suggested adjustments?

CM000n: ping @AndyVRD ;-)

CM000n (approved):

CM000n:
I think I'll merge for now and tackle the suggested changes as code improvements later.

checkerschaf:
I've fixed the issue in #423 👍

PR #369, opened 2022-01-31
Different take on card documentation. 

stokkie90:
Screenshot from 2022-01-31 22-48-32
Screenshot from 2022-01-31 22-48-21


CM000n:
I like it :-) But IMHO there are a bit too many unnecessary icons and colours that distract from the essentials. Maybe we can just use white or/and the corresponding dark theme colour for all admonitions here?
It is also kind of redundant to have the name of the card both in a heading and in the header of the Admonition Box.

data/card_templates.yaml
```diff
@@ -0,0 +1,37 @@
+---
+card_templates:
```
CM000n, line 2: Can you explain a little more what this `card_templates.yaml` is for and how it works?
I saw that you reference it in the macros in `mkdocs.yml`.

stokkie90: Yeah, was my first idea. to use a yaml data file as source to create a MD file. 

But i maybe found something more interesting. if you check the `main.py` file in the root of the project. 

it is some python code that will search all the yaml in ulm_template dir. and will check if the yamls have a `doc:` part. If so it will be added to a jinja variable which i can use to loop over it. (see card_list.md)

macros module: https://mkdocs-macros-plugin.readthedocs.io/en/latest/macros/

mkdocs macros allows you to do jinja2 magic inside a md file!

stokkie90:
> I like it :-) But IMHO there are a bit too many unnecessary icons and colours that distract from the essentials. Maybe we can just use white or/and the corresponding dark theme colour for all admonitions here? It is also kind of redundant to have the name of the card both in a heading and in the header of the Admonition Box.

We can off-course use other abmonition: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#configuration or even custom: https://squidfunk.github.io/mkdocs-material/reference/admonitions/#custom-admonitions

wilbiev:

Issue #386, opened 2022-02-04
Better explanation of how to add the existing custom cards


sildehoop:
**Is your enhancement request related to a problem? Please describe.**
Some people find it unclear how to add the existing custom cards #384

**Describe the solution you'd like**
Better description of how the add the existing custom cards

CM000n:
Cloesed with https://github.com/UI-Lovelace-Minimalist/UI/pull/387

Issue #384, opened 2022-02-04
"Button-card template is missing" error


RakhmatovR:
Hi there! Just installed the integration via HACS and currently trying to add some controls. Faced with such error while using custom cards:
```
Button-card template 'custom_card_tpx01_aircondition_with_buttons' is missing!

type: custom:button-card
template: custom_card_tpx01_aircondition_with_buttons
variables:
  entity: fan.bedroom_humidifier
  name: Humidifier
```

Shouldn't them be included already into integration?

sildehoop:
Follow the instructions on https://ui-lovelace-minimalist.github.io/UI/setup/configuration/#custom-cards

RakhmatovR:
@sildehoop Does it mean that I should copy customs cards from repository into the `custom_cards` folder?

sildehoop:
Yes

RakhmatovR:
@sildehoop Ok, thanks. I think this should be documented somewhere because it's not so obvious.

PR #387, opened 2022-02-04
Improve custom card documentation

basbruss:


sildehoop:
If you want you can add #386 ??

basbruss:
@sildehoop  that's more related to the usage  of custom_cards instead of developing them right? I can try to improve the docs for the usage of custom cards

sildehoop:
Yes I know but #384, yes maybe that is the solution

basbruss:
@sildehoop I have created a new page for downloading custom_cards and referenced from the old place to the new page

sildehoop (dismissed):

mkdocs.yml
```diff
@@ -15,6 +15,7 @@ nav:
       - Download: "setup/download.md"
       - Installation: "setup/installation.md"
       - Configuration: "setup/configuration.md"
+      - Custom_cards: "setup/custom_cards.md"
```
basbruss, line 18: @stokkie90 Is this the only change I need to do to create a new page?

basbruss, line 18: @stokkie90 Is this the only change I need to do to create a new page?

stokkie90: The key in this case is the Title. So i would change it to: `"Custom Cards": "setup/custom-cards.md"`

stokkie90 (approved):

PR #389, opened 2022-02-04
Add Czech & Slovak translations.

mstefany:
This PR adds Czech and Slovak translations to UI Lovelace Minimalist. Most of the strings were translated according to how they are translated in Home Assistant Core. The rest was translated with the best intent. Both translations were checked by another person, so they should be correct.
Some strings, e.g. on or off, can be translated to more forms, depending on the object described - zapnutý/zapnutá/zapnuté - so the proposed form might need to be changed in the future if its usage shows that it's currently not good form. I am still a very new user of this UI.
Thank you.


CM000n:
Hello, you submitted your pull request as a draft.
Do you plan to make any changes to your PR @mstefany?
If not, we could remove the draft label and merge the PR.

mstefany:
Hello @CM000n, I just needed a bit more time to finish the translation and compare it to the translated strings in Home Assistant Core. Now I marked it for review. I would be very happy if this gets included in version 0.0.5.
Thank you for this great UI project!

CM000n (approved):

custom_components/ui_lovelace_minimalist/configure.py
```diff
@@ -15,6 +15,7 @@
 _LOGGER: logging.Logger = logging.getLogger(__name__)
 
 LANGUAGES = {
+    "Czech": "CS",
```
CM000n, line 18: Please forgive my ignorance. But shouldn't the ISO country code for the Czech Republic be "`CZ`" instead of "`CS`"?

custom_components/ui_lovelace_minimalist/lovelace/translations/CS.yaml
```diff
@@ -0,0 +1,40 @@
+---
```
CM000n, line 1: Also here in the filename. "`CZ`" instead of "`CS`"?

custom_components/ui_lovelace_minimalist/translations/cs.json
```diff
@@ -0,0 +1,43 @@
+{
```
CM000n, line 1: Also here in the filename. "`CZ`" instead of "`CS`"?

CM000n (changes requested):

custom_components/ui_lovelace_minimalist/configure.py
CM000n: Interesting, you're right. 
But now we have a mixture of different logics. Either you should use the ISO code of the country and not the locale one for the time being, or we should change everything over to it.
What do you think about it @stokkie90?

stokkie90: According to the home assistant [docs](https://developers.home-assistant.io/docs/translations) And looking at this page. Both are possible? https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry

```
Type: region
Subtag: CZ
Description: Czechia
Description: Czech Republic
Added: 2005-10-16
......
Type: language
Subtag: cs
Description: Czech
Added: 2005-10-16
Suppress-Script: Latn
```
But the language is indeed in this case `CS`

CM000n (approved):

PR #390, opened 2022-02-06
Custom Person Card + Popup

imswel:
I have created a custom card that improve the `card_person` by allowing the user to set up as many zones as desired and adding a popup for the card inspired by the [@matt8707](https://github.com/matt8707/hass-config)'s theme.

popup


CM000n:
This looks great @imswel . Thanks a lot!
Have you ever thought about adapting the integrated people card accordingly instead of making a custom component for it?
I think in this case it would fit quite well with the rest of the theme. Maybe @schumijo has an opinion on the popup?

schumijo:
Hi @imswel @CM000n 

Good job. Love the structure of your popup. But I'm not sure it fits the rest of the popup theme with 7ahang's inspiration. Maybe few small tweaks on sensor display and button also are necessary. 
It's just my opinion 😉

schumijo:
And also more important : you should split your card and your popup in specific yaml files

CM000n:
Ok, so we should add it as a custom_card for now, and maybe think about it later if and which features we want to integrate into our standard person card.

CM000n (approved):

CM000n (approved):

PR #392, opened 2022-02-08
Hide slider if light is unavailable

drthanwho:
If the ligh card has the slider enabled and `ulm_card_light_enable_collapse: true` it's supposed to be hidden when the light is off. However if the light is unavailable the slider will show up. 

image

This addes a condition to acount for entity being unavailable and hide the slider as it does when off
image

I'm by no means an expert in teplates. This works but perhaps there's a better way to do it 😅


basbruss:
This should already be fixed with #375

drthanwho:
Ah indeed I didn't notice the dev branch. Will close this

PR #397, opened 2022-02-10
[custom_card_homeassistant_updates] Translation and Credits Update

AndyVRD:
- Updated translations

PR #400, opened 2022-02-12
Language update on Custom card homeassistant_updates

r0binj:
Corrected the language so it isn't mixed :)


r0binj:
Woops, missed the corrections from PR: [#397](https://github.com/UI-Lovelace-Minimalist/UI/pull/397)
My bad :)

PR #409, opened 2022-02-14
Added tap_action to chip_icon_double_state.yaml

LRvdLinden:
```yaml
  tap_action:
    action: "navigate"
    navigation_path: "[[[ return variables.ulm_chip_navigate_path; ]]]"
```


CM000n:
Hello, thank you very much for your contribution! That's a good idea with the navigation path.
Unfortunately, I think your PR is a bit incomplete and inconsistent.
- You should also adapt the documentation accordingly when you add a new functionality somewhere.
- Why do you limit your change only to the `chip_icon_double_state`? Isn't that a bit inconsistent? Shouldn't we add this functionality to the rest of the state chips as well?

I was also wondering what happens if no `ulm_chip_navigate_path` is specified? Does nothing happen when you tap on the chip, or does an error message appear?

basbruss (approved):

basbruss (approved):

PR #423, opened 2022-02-16
Fix and improve card_battery

checkerschaf:
This fixes the issue that iOS devices report their battery_state as `Charging` instead of `charging` by using `.toLowerCase()` on the state.

Additionally I've made some changes to the code to improve the code style a bit 😄


CM000n:
Mää, that looks great. 
Thank you very much!

CM000n (approved):

PR #434, opened 2022-02-18
Card Media Player:

leon38:
- Get default icon when no app in media_player attributes
- Get media player friendly name when no media title is set


custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_media_player.yaml
```diff
@@ -36,7 +38,14 @@ card_media_player:
   state:
     - operator: "template"
       value: "[[[ return entity.state != 'off' ]]]"
-      name: "[[[ return states[entity.entity_id].attributes.media_title; ]]]"
+      name: |
+        [[[
+          var name = states[entity.entity_id].attributes.friendly_name;
+          if(states[entity.entity_id].attributes.media_title){
+            var name = states[entity.entity_id].attributes.media_title;
```
kubiksamek, line 45: JS Nitpick: No need to use `var name` again. Also, it would be better to use `let name` in the first place.

kubiksamek (approved):

CM000n:
You changes have some Merge Conflicts with previous changes.
Can you please have another look?

CM000n (approved):

PR #449, opened 2022-02-20
thermostat label off mode added

eraycetinay:
generic_thermostat includes 'off' mode.
Ekran Resmi 2022-02-20 17 04 39

but in label there are heating, cooling and idle options.  so it shows 'unavailable' for my devices. it shouldn't.

Ekran Resmi 2022-02-20 17 04 58

i firstly added 'off' condition to 'mode'  but then i realized a device cant be in a heating, cooling or idle mode if it is off.  
so i disabled 'mode' if device entity state is off.  i think this look better.


eraycetinay:
also i need to send another commit for my programmable_thermostat.  because it always show 'off' on hvac_action attribute but its state is heat or cool.  so im not sure about the best practice here but, maybe we can fully disable the unavailable state . this is probably another topic but i need a fix on this too. im just not figured out yet.

CM000n (approved):

PR #443, opened 2022-02-18
[bugfix] Changed some Toolbar sizing and padding in Tapbar theme

Ned-Tom:
Hopefully this fixes multiple issues on the iPhone mentioned in #346 and on the HA Forums : https://community.home-assistant.io/t/lovelace-ui-minimalist/322687/1522

## Additions
- app-header-selection-bar-color variable to hide a black bar on Android.
- some color consistency changes to make the theme more uniform.

## Needs fixing
- [x] Ios (Iphone) tapbar bottom spacing and overlap.
- [x] view top shifting when loading and switching views
- [x] Tapbar jerking when scrolling to the end.

## Need help and feedback
The implementation of the tapbar as requested #346 is not that easy and brings some bugs whit it.
If you have an Idea Pleas suggest.

_Don't post as fix because it can break things on other devices._

PR #453, opened 2022-02-21
Little more headroom when using the tapbar and swipe-card

bruvv:
Also when you are using this awesome [swipe-card](https://github.com/bramkragten/swipe-card) it needs a little more room at the bottom.
Without fix:
image
With fix:
image


bruvv:
Oeps duplicated: https://github.com/UI-Lovelace-Minimalist/UI/pull/443

bruvv:
This is not how it is done I see. I will make an issue about this.

PR #460, opened 2022-02-22
Use variables instead of sensor values to set colors for Afvalophaling

Tommatheussen:
I tried adding this card to my dashboard, and it kept failing because the code referenced sensor values instead of using the defined variables.


AndyVRD:
Thanks, forgot to change wen i pushed the PR.

CM000n (approved):

PR #440, opened 2022-02-18
Format Card Titles for Wiki

basbruss:

PR #464, opened 2022-02-23
Custom card device tracker

aehru:


basbruss (changes requested):
Could you change the title to be inline with the title format from #440 ?

aehru:
> Could you change the title in the readme.md to be inline with the title format from #440 ?

Oh yes, didn't saw that, will do it.

basbruss (approved):

basbruss:
What are the appropriate translations? as some are direct translations of `online` and `offline`, and some `home` and `not_home`

aehru:
This require more translations for all languages. Or I can apply same translations as 'home' and 'not_home' as I started to do but was not sure what is best option.

CM000n:
The best would be to use "Online" and "Offline" as placeholders if you don't know the correct translation.
In German it would be wrong to translate a device as "Zu Hause" if I want to express its power-on or connection status. Unless you are talking about mobile devices and want to describe geographically whether they are at home or not ;-)

aehru:
The purpose is to use it for mobile or non mobile devices so I will update all with placeholders as suggested.

basbruss:
> The best would be to use "Online" and "Offline" as placeholders if you don't know the correct translation. In German it would be wrong to translate a device as "Zu Hause" if I want to express its power-on or connection status. Unless you are talking about mobile devices and want to describe geographically whether they are at home or not ;-)

I agree with this. But when you want to describe their geographically state. It's better to use the already existing `ulm_person_state_*` translations, to avoid over-cluttering of the translations files.

basbruss (approved):

basbruss (approved):

PR #465, opened 2022-02-23
Add usb charging option for card_battery

aehru:


CM000n:
Hi @aehru, it seems you developed against the wrong branch an therefore now you MR have some Merge Conflicts.
Can you please have another look at it?

aehru:
Oh yes, I developed on the dev branch on my side but it seems to have disappear on the main repo. So for the future I will need to use the release branch?

CM000n:
> Oh yes, I developed on the dev branch on my side but it seems to have disappear on the main repo. So for the future I will need to use the release branch?

It depends... ;-)
Things like custom cards or cosmetic changes in the wiki that are not directly related to code changes can be developed directly against the Main Branch.
Other things that affect the custom component for HACS should always be developed against the release branch.
This is admittedly a bit confusing. In the future, we will have to think about how we can better design our branches and github actions.

CM000n (approved):

PR #484, opened 2022-02-28
fix(card/thermostat): use hass configured system temperature unit

ulic75:
Two issues with commit https://github.com/UI-Lovelace-Minimalist/UI/commit/54157772e32796f64e66be287fbdc80c34935820:

1. `var temp_unit = entity.attributes.temperature_unit;` isn't defined in the top chunk, so I suspect that would be an issue
2. My thermostat (Nest) doesn't have an attribute named `temperature_unit`, I suspect others may not as well.

The proposed change uses the system configured unit, which is how the official thermostat card in Home Assistant does it.


CM000n:
Thank you for your contribution. Our current branch layout is indeed a bit confusing. Revising this is also on our to-do list.

CM000n (approved):

PR #500, opened 2022-03-02
Use entity instead of variable where possible to be consistent

### LINKED ISSUE #478: Graph card does not use entity name by default
**Describe the bug**
The graph card does not use the entity name by default. It just shows `n/a` if nothing is specified.

**To Reproduce**
1. Create a graph card:
2. Don't set `ulm_card_graph_name`

**Expected behavior**
Take the entity name by default.

**Screenshots**
image

**Additional context**
```yaml
- type: 'custom:button-card'
  template: card_graph
  variables:
    ulm_card_graph_entity: sensor.temperature_inside
    ulm_card_graph_type: fill
    ulm_card_graph_hours: 24
```

TheZoker:
Almost all cards use the main "entity" attribute to set the entity that should be used. 
When I tried different cards, I noticed, that some are using a ulm_*+entity variable instead. 

To be consistent and allow to easily switch between cards, I adjusted the cards to use the entity attribute. This is a breaking change and I'm not sure, how to communicate this properly. 

Fixes #478

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/2-line_cards/card_graph.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/2-line_cards/card_graph.yaml
@@ -1,9 +1,10 @@
 ---
 ### Card Graph ###
 card_graph:
+  show_name: false
+  show_icon: false
   variables:
     ulm_card_graph_color: "var(--info-color)"
-    ulm_card_graph_name: "n/a"
     ulm_card_graph_color2: "var(--info-color)"
     ulm_card_graph_name2: "n/a"
     ulm_card_graph_entity2: ""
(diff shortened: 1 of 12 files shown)


custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/2-line_cards/card_graph.yaml
```diff
@@ -31,15 +32,15 @@ card_graph:
             - box-shadow: "none"
             - border-radius: "var(--border-radius) var(--border-radius) 0px 0px"
             - padding: "12px"
-        entity: "[[[ return variables.ulm_card_graph_entity ]]]"
```
basbruss, line 34: When multiple entities can be assigned, I think it's better to store them all in variables like (variables.entity_1; variables.entity_2; etc) to avoid confusing that one of them is not set in the variables but directly to `entity:`

TheZoker: In general I agree with you for something like `chip_icon_double_state`, but I think that most of the time, the graph card will be used with only one entity. But I can revert this change if it is not wanted :)

TheZoker:
I changed the target branch to `release` and also applied this to the chip cards

rairulyle:
Would love to get this merged! This would also allows us to easily use auto-entities!

PR #501, opened 2022-03-02
Add a custom icon variable for the custom room script

TheZoker:
Currently the default icon for the custom room script button is used.

With the new variable the user can set his own custom icon (also the new default icon is the sofa icon, just like the screenshots).

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_vacuum.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_vacuum.yaml
@@ -9,6 +9,7 @@ card_vacuum:
     ulm_card_vacuum_name: "[[[ return entity.attributes.friendly_name ]]]"
     ulm_card_vacuum_icon: "[[[ return entity.attributes.icon ]]]"
     ulm_card_vacuum_room:
+    ulm_card_vacuum_room_icon: "[[[ return entity.attributes.icon ]]]"
   triggers_update: "all"
   styles:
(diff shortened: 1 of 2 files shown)


custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_vacuum.yaml
```diff
@@ -9,6 +9,7 @@ card_vacuum:
     ulm_card_vacuum_name: "[[[ return entity.attributes.friendly_name ]]]"
     ulm_card_vacuum_icon: "[[[ return entity.attributes.icon ]]]"
     ulm_card_vacuum_room:
+    ulm_card_vacuum_room_icon: "mdi:sofa"
```
basbruss, line 12: I do agree with adding the option to change the icon manually. But I think de default should be the icon set in the GUI of HA. As this will now always be overridden by the default.
This can be achieved with 
```icon: "[[[ return entity.attributes.icon ]]]"``` or 
` icon: "[[[ return states[variables.ulm_card_vacuum_room].attributes.icon ]]]"`

basbruss:
Hi @TheZoker 

Thanks for your contributions to make this theme even more awesome!
I have added some remarks/suggestions to your alterations. Could you take a look at these?
Also when a PR does not only include changes in readme or custom_cards files it should be directed to `release` branch instead of the `main` branch. 😉

TheZoker:
Hi @basbruss 
Thanks for the review! I changed the base branch to `release` and updated the default icon behavior!

PR #514, opened 2022-03-07
fix(customcard/ha_updates): capitalize OS

ulic75:
Typically when the first letters of words are used as an acronym that do not create a new word they are all capitalized.  For example FBI, no one says fibi, they say the letters, F, B, I.  Same applies for OS (at least in English).

Tagging @AndyVRD at the card creator.

Feel free to reject if you don't like this change.


AndyVRD:
Good catch. 👍🏻

basbruss:
@ulic75 could you change the branch to `main`? Changes to custom cards are not release dependent and therefore directed to the `main` branch 😉 . When I tried to do it for you it adds some unwanted changes to this PR.

ulic75:
Will resubmit against main.

PR #534, opened 2022-03-15
[popup_card_light] add color to slider

basbruss:
Idea by @AndyVRD 
Add Light Color to brightness slider
Checks for rgb_color attribute like `card_light` does
before:
image
after:
image


schumijo (approved):
Good idea !

PR #542, opened 2022-03-19
card_media_player: Enhance icon, title and label by adding artist name and album name at the same time

rairulyle:
_NOTE: Size does not change_

**Before:**
image

**After:**
2022-03-10 02_10_23-Clipboard


custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_media_player.yaml
```diff
@@ -97,7 +97,7 @@ card_media_player:
                 ulm_popup_media_player_entity: "[[[ return entity.entity_id ]]]"
         icon: |
           [[[
-            var icon = "mdi:speaker";
+            var icon = states[entity.entity_id].attributes.icon;
```
mazerte, line 100: This is useless because 3 "if" below you already have the case but with the change we are loosing the default value

mazerte (commented):
I'm not sure everybody want this behaviour. I personal use the card in a small size and adding `media_artist` won't fit on my case. Maybe you can add it as a variable

custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_media_player.yaml
rairulyle: This line is for custom players and it means that it will use the entity/attribute icon automatically instead of a constant value"mdi:speaker". By default, HomeAssistant also set it to "mdi:speaker" so no issue with that. It just gives the option to be dynamic.

rairulyle: > I'm not sure everybody want this behaviour. I personal use the card in a small size and adding `media_artist` won't fit on my case. Maybe you can add it as a variable

Will do try to make it an option :)

basbruss:
Thanks for you contribution! All requests to non-custom_cards should be allocated to the release branch 😉 
Could you redirect this PR?

rairulyle:
> Thanks for you contribution! All requests to non-custom_cards should be allocated to the release branch 😉 Could you redirect this PR? Making things optional is a better approach as mentioned by @mazerte

I'll close the pr and submit a PR again. It's getting unnecessary changes when I PR it to release.

PR #543, opened 2022-03-20
card_media_player: Enhance icon, title and label

rairulyle:
**Fix #542 PR**

_NOTE: Size does not change_

**Before:**
image

**After:**
By setting `ulm_card_media_player_more_info` to `true` it will show the following:

**SMALL**
image

**LARGE**
image


rairulyle:
@basbruss kindly help check again. Will make a change to make it optional soon.

rairulyle:
@basbruss @mazerte ,

I made a change. Now it gets the `entity.attributes.icon` by default if it is available and if it is not, it is set to `"mdi:speaker"`. Works even when the device is not playing anything!

I also updated the images above. I think we don't need to have an option for short/full song information as the new button-card cuts the text and does not overflow. Let me know if you guys still want the option.

basbruss:
@rairulyle I will try tomorrow your changes to check if the option is needed.

mazerte:
Hi @rairulyle thanks for this. I still think that having the `more_info` parameter is important because on small version is very likely to cut and isn't very visually pleasing to have that. I don't think a cost a lot to add this option and it will give more flexibility to the user.

rairulyle:
> Hi @rairulyle thanks for this. I still think that having the `more_info` parameter is important because on small version is very likely to cut and isn't very visually pleasing to have that. I don't think a cost a lot to add this option and it will give more flexibility to the user.

Noted on this @mazerte ! Will be adding that option.

davidepanato (dismissed):

finder39:
@rairulyle the correction for the `variables.ulm_card_media_player_name` name setting is clutch. Thanks!

basbruss:
@rairulyle Did you already had some time to implement this?
> Noted on this @mazerte ! Will be adding that option.

rairulyle:
Forgot this one! I'll commit my change tonight :)

> @rairulyle Did you already had some time to implement this?
> 
> > Noted on this @mazerte ! Will be adding that option.

rairulyle:
Hi @basbruss , done!

basbruss:
@rairulyle Thanks! 
Is your latest commit supposed to delete the option to have a volume slider? And could you also a the new variable to the documentation/readme file?

rairulyle:
@basbruss , let me submit a new one. It got a conflict and messed up with the latest code in the release branch. Also added in the documentation already.

basbruss (approved):

PR #567, opened 2022-03-31
Allow today or tomorrow collections to be optional for AfvalOphaling

Tommatheussen:
A couple of updates to the custom card AfvalOphaling:

- some small speed improvements (probably negligible) when rendering the labels
- Check for the existence of  `ulm_card_ophaling_vandaag` and `ulm_card_ophaling_morgen` when trying to load states for them. In my case, the garbage collection happens in the early morning, way before I get up in the morning so I don't use the 'Today' sensor at all, we put our bins outside the night before. On top of that, collections usually happen 2 days in a row, meaning when the Today is being shown,  I don't see the tomorrow collection and forget to put it out :)

### DIFF
--- a/custom_cards/custom_card_afvalophaling/card_afvalophaling.yaml
+++ b/custom_cards/custom_card_afvalophaling/card_afvalophaling.yaml
@@ -16,6 +16,14 @@ card_afvalophaling:
         ]]]
   label: |
         [[[
+          if(states[variables.ulm_card_ophaling_vandaag].state !=='Geen'){
+            return states[variables.ulm_card_ophaling_vandaag].state
+          }
+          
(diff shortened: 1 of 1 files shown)


basbruss:
@AndyVRD could you check the proposed changes?

AndyVRD:
I will check it tomorrow, i think the changes make sense.

Note: The card was actually build for personal use and the trash collections for my village.

basbruss (approved):

basbruss:
@Tommatheussen Could you also make a new PR to bring the documentation inline with the changes?
I accidentally merged it before checking on documentation

Tommatheussen:
Sure thing! I just created the PR: https://github.com/UI-Lovelace-Minimalist/UI/pull/584
let me know if there's anything else that needs to added :)

PR #570, opened 2022-04-01
Fix april fools day inclusion 

stokkie90:


basbruss (approved):

thomasloven:
I'm sorry you got caught in my mistake as well.

basbruss:
> I'm sorry you got caught in my mistake as well.

No need to worry! I personally can appreciate the joke 😄 
I also agree with the underlying message and think we do need to make users more aware of possible vulnerabilities in the scripts they install for their ease of use.

PR #571, opened 2022-04-01
Bump card mod

SNoof85:
Disable april fool 🍡
See : https://github.com/thomasloven/lovelace-card-mod/releases/tag/3.1.4


basbruss:
Thanks!

The fix is already online 😄 
https://github.com/UI-Lovelace-Minimalist/UI/releases/tag/v0.0.7-hotfix2

PR #582, opened 2022-04-05
Update minimalist-ios-tapbar.yaml

LRvdLinden:
## Bug fix
---
- [x] Non-Minimalist cards became squared
- [x] The tapbar touches iOS navigation bar, while having a lot of margin-top
- [x] The tapbar sometimes (after refresh) was higher than expected, and it resized on tap, jumping a bit.

Thanks to [antonio1475](https://github.com/antonio1475) for helping out!


basbruss:
@LRvdLinden Thanks for the quick fixes. Could you redirect the PR to the `release` branch?

PR #588, opened 2022-04-07
Fix pre-commit stuff + bump repo versions

basbruss:
@stokkie90 @CM000n  Could you guys check the changes make sense? I updated all repo versions I could find due to some errors the pre-commit action had the last week.
Isn't an idea to use something like dependabot to automate this process?


stokkie90:
i agree we should look into something like dependabot for some dependencies indeed.

stokkie90 (approved):

PR #592, opened 2022-04-09
Add generic collapsible option for media controls and volume slider

rairulyle:
Collapsible is a generic feature in most cards so I added it in the media player too! This is to ensure that spaces are not wasted when cards are not used.

**BEFORE:**

State: OFF
CURRENT_OFF

State: PLAYING - _(No Change)_
AFTER_PLAYING

**AFTER**

State: OFF
AFTER_OFF

State: PLAYING - _(No Change)_
AFTER_PLAYING


rairulyle:
@basbruss kindly help review :)

basbruss (approved):

PR #603, opened 2022-04-11
custom_card_esh_welcome - Weather Chip fixes

AndyVRD:


basbruss (approved):

josheharvey:
Card does not display. No error's thrown by the system. Should names be "" or not?

PR #610, opened 2022-04-13
Added Brazilian Portuguese translation

kauelima:
There was already a pt translation which pt-BR users were fallback to, but there are some key differences.


basbruss:
@kauelima Thanks for the translations!

I have pushed a commit with the needed python config to load the language-file in the integration 😄

basbruss (approved):

kauelima:
> @kauelima Thanks for the translations!
> 
> I have pushed a commit with the needed python config to load the language-file in the integration 😄

Thank you a lot! I did not know I needed to add this call. :)

PR #621, opened 2022-04-19
Add garage and gate types to cover card

yagrasdemonde:
Same way as for the curtains, I added garage and gate as types of cover (with the open/closed icons) to cover all covers possible.


yagrasdemonde (commented):
Fix typo in navigation path for fifth entity

basbruss (approved):

PR #648, opened 2022-05-06
Custom card more power outlet

wranglatang:
This is a new custom card that provides additional information with regards to a power outlet. Quite often I have wanted to know how much energy has been consumed vs how long something has actually been on.


wranglatang:
I have corrected the white space issue.

PR #684, opened 2022-05-22
Added a new Custom Door Lock card

Gilganik:
Added a new Custom Door Lock card


Gilganik:
files need to be changed

PR #694, opened 2022-05-25
Adjusted custom lock card

13robin37:
Added the option to only use the card to open a lock.
Also added "jammed" state translation

### DIFF
--- a/custom_cards/custom_card_eraycetinay_lock/README.md
+++ b/custom_cards/custom_card_eraycetinay_lock/README.md
@@ -7,7 +7,7 @@ hide:
 
 # Custom-card "Lock"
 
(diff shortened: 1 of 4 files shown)


13robin37:
@Gilganik we might want to merge the two lock cards if you're okay with it. Maybe an option to switch between this layout (one button) and your layout (two buttons)? 😊

Gilganik:
> @Gilganik we might want to merge the two lock cards if you're okay with it. Maybe an option to switch between this layout (one button) and your layout (two buttons)? 😊

To me it's fine, what should I do?

13robin37:
@Gilganik I should have some time in the next few days, will test and open a PR once I merged them.

Gilganik:
> @Gilganik I should have some time in the next few days, will test and open a PR once I merged them. 

Fine for me! I think it will be great

basbruss (approved):

PR #695, opened 2022-05-25
Adjusted vacuum card

13robin37:
Added a camera entity and the option to only show it during cleaning.


basbruss (approved):

basbruss:
@13robin37 Nice addition! 😄

PR #713, opened 2022-05-31
Fix language variables

Hoellenwesen:
The other languages didn't work because he got confused with the variables. The language variables are now without "ulm_" and are loaded properly.

### DIFF
--- a/custom_cards/custom_card_yagrasdemonde_lights_count/custom_card_yagrasdemonde_lights_count.yaml
+++ b/custom_cards/custom_card_yagrasdemonde_lights_count/custom_card_yagrasdemonde_lights_count.yaml
@@ -3,7 +3,7 @@
 custom_card_yagrasdemonde_lights_count:
   show_name: true
   template:
-    - "ulm_custom_card_yagrasdemonde_lights_count_language_variables"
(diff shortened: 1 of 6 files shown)


basbruss:
Hi @Hoellenwesen, 

Why are these changes needed? Changing variable names does nothing to how it is supposed to work.
Also as far I am concerned the card works as it should.

Hoellenwesen:
Hi,
i just tried something. I used the card and it was not working, some error with missing variables. Then i played around and renaming those variable worked for me, after these changes, the card worked on my dashboard. 

If you say it is not needed then maybe i have done something wrong, but i don't know what it is. For me i will keep these changes locally, because it is working.

basbruss:
That's probably an issue with multiple files with the same language variables template.
Check `custom_components/ui_lovelace_minimalist/__ui_minimalist__/custom_cards/custom_card_yagrasdemonde_lights_count` for multiple languages files. The variables only work if there is just one language file. You can safely remove unwanted ones.

Hoellenwesen:
Hi @basbruss,
i have tested this. I used the code from the github repo and deleted all language files except for the DE.yaml. I reloaded everything but the card is still in english.

If i change the variables as i did before, without the "ulm_" then the card is in german.

PR #718, opened 2022-06-01
Added new Custom Clock Card

Gilganik:
Added new Custom Clock Card


custom_cards/custom_card_nik_clock/custom_card_nik_clock.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - "ulm_language_variables"
+  variables:
+    ulm_custom_card_nik_clock_label: "[[[ return variables.ulm_custom_card_nik_clock_label ]]]"
+    ulm_custom_card_nik_clock_switch: "[[[ return variables.ulm_custom_card_nik_clock_switch ]]]"
+    ulm_custom_card_nik_clock_switch_enable: false
+  show_icon: false
+  show_name: false
+  show_state: true
+  show_label: true
+  entity: "sensor.time"
```
basbruss, line 13: Hardcoding an entity is now advised as not everyone has set up this entity.
You could use some Javascript template to extract the local time.
```suggestion
  name: >
    [[[
      const time = new Date();
      let hour = time.getHours();
      let minute = time.getMinutes();
      return hour + ":" + minute;
    ]]]
```

Gilganik:
@basbruss many thanks for the advices, I've applied them all. One question, the Minutes shows just one number. Like this: "16:7", insted of "16:07" do you know how to address it?

PR #719, opened 2022-06-01
Added Custom clock Card

Gilganik:
@basbruss  Thanks for your suggestion. I've found out how to fix the 2 numbers minutes :)
Let me know if now is ok


Gilganik:
@basbruss  it should be ok with the changes? let me know your feedback

basbruss (approved):

PR #725, opened 2022-06-04
Update to fix missing variables

modem158:
1. Changed variables in the name area to strings.  The variables were never set and don't need to exist.
2. Changed the logic in the name area to support graphs less than 1 hour.

### DIFF
--- a/custom_cards/custom_card_damix48_power_details/custom_card_damix48_power_details.yaml
+++ b/custom_cards/custom_card_damix48_power_details/custom_card_damix48_power_details.yaml
@@ -32,10 +32,10 @@ custom_card_damix48_power_details:
         entity: "[[[ return variables.ulm_card_power_details_entity ]]]"
         name: |
           [[[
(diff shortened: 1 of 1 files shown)


Eltarius:
This PR is not necessary

> 1. Changed variables in the name area to strings.  The variables were never set and don't need to exist.

See https://github.com/UI-Lovelace-Minimalist/UI/tree/main/custom_cards/custom_card_damix48_power_details/languages
And https://ui-lovelace-minimalist.github.io/UI/development/custom_cards/ to use language files 😉

basbruss:
Hardcoding in English strings is not recommended and is already provided by translation variables as mentioned by @Eltarius .
Changing the logic seems fine to me.

modem158:
Thanks you two.  I was unaware of how the language files work and a little too eager to contribute.  

Should I just close the pull request?

modem158:
Closing this pull request as I wasn't aware of how the en.yaml languages worked making this update isn't necessary.

PR #726, opened 2022-06-05
Update it translation

Petapton:


basbruss:
Adapted by #761 to new translation engine

PR #728, opened 2022-06-05
Added translation engine

13robin37:
Added universal translation states and adjusted states for most cards.
Still some small tweaks to fix.

This should avoid PR #727 and #726. We've discussed a few things in #711. Will update this with more commits during the next few days, then most cards should support all translated languages.


13robin37:
Okay, won't avoid #727, I cannot fetch those strings for now. I've modified the DE translation file to only reflect strings that aren't present in the hass object and will do so for the other files with a coming PR once everything is behaving like it should.

I've tested the card_generic, card_generic_swap, card_light, card_media_player, card_person, card_power_outlet and card_vacuum.

Still needs to be tested:
- card_battery
- card_cover
- card_input_boolean
- card_room
- card_thermostat (especially this one)

I've adjusted the translations for the popups as well, most of them are available in HA. 

Modified the documentation for custom_cards to reflect the change.

We should evaluate whether it's necessary to supply the person entity as a variable for the person cards. In my opinion it would result in cleaner code to just supply it as an entity as you could just use the ulm_translation_state variable in that case. Currently I've switched it to a shorthand form of the translation variable. This would be a breaking change though.

basbruss:
> We should evaluate whether it's necessary to supply the person entity as a variable for the person cards. In my opinion it would result in cleaner code to just supply it as an entity as you could just use the ulm_translation_state variable in that case. Currently I've switched it to a shorthand form of the translation variable. This would be a breaking change though.

We can add a line of code to the person card to add the entity without breaking existing configurations. Maybe add a warning that the old way in variables is not advised anymore and will be deprecated after some time.
```yaml
entity: "[[[ return variables.ulm_card_person_entity ]]]"
```

It will also take some time to test this pr in my dev-env

13robin37:
> ```yaml
> entity: "[[[ return variables.ulm_card_person_entity ]]]"
> ```

I've tested this and was unable to make it work, when I removed the `entity: "[[[ return variables.ulm_card_person_entity; ]]]"` from tap_action it wasn't working anymore. Maybe someone with more patience will be able to make it work.

basbruss:
I didn't realize that line was already in the template 😅 
I also don't get it to work without setting the entity manual.

Converting everting to entity instead of a variable seems to be better. So introducing a breaking change doesn't seem to be problem I think.

schumijo:
It's an awesome improvement guys ! Thank you

custom_components/ui_lovelace_minimalist/lovelace/translations/default.yaml
```diff
@@ -0,0 +1,104 @@
...
+      return(hour);
+    ulm_translation_hours: >
+      var number = hass.resources[hass['language']]["ui.duration.hour"];
+      var matches = number.match(/[^{\}]+(?=})/g);
+      var hour = "Hours";
+      if (matches) {
+        var hour = matches[2];
+      }
+      return(hour);
+    ulm_translation_hvac: >
```
basbruss, line 24: This was not working for me
I've adjusted the code a bit an now it works as expected
```suggestion
    ulm_translation_hvac: >
      [[[
        if (typeof(entity) !== 'undefined' && entity !== undefined){
          let state = entity.state;
          let d_class = entity.attributes.device_class;
          let def = ["unknown", "unavailable"];
          let lang = hass["language"];
          let action = entity.attributes.hvac_action;
          let domain = entity.entity_id.substr(0, entity.entity_id.indexOf("."));
          var mode = hass.resources[lang]["state_attributes." + domain + ".hvac_action." + action];
          if (def.includes(state)) {
            var mode = hass.resources[lang]["state.default." + state ];
          }
          var label = hass.resources[lang]["component." + domain + ".state._." + state] || state;
          return (entity.attributes.current_temperature ) + '°' + ' • ' + label + (entity.state !='off' ? ' (' + mode + ')' : '');
          }
      ]]]
```

custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_light.yaml
```diff
@@ -200,17 +200,17 @@ card_light:
                 [[[
```
basbruss, line 200: This can be more efficient
```suggestion
                [[[
                  var label = variables.ulm_translation_state
                    if (entity.attributes.brightness != null && entity.state === 'on'){
                      var bri = Math.round(entity.attributes.brightness / 2.55);
                      var label = (bri ? bri : "0") + "%";
                    }
                  return label;
                ]]]
```

custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_person.yaml
```diff
@@ -21,14 +21,14 @@ card_person:
         return states[variables.ulm_address].state
       } else {
         let state = states[variables.ulm_card_person_entity].state;
-        return variables["ulm_person_state_" + state] ? variables["ulm_person_state_" + state] : state;
+        return (state == "home" || state == "not_home") ? hass.resources[hass["language"]]["component.person" + ".state._." + state] : state == "unavailable" ? variables.ulm_translation_unavailable : state;
```
basbruss, line 24: Changed logic to also give option for `unknown`
```suggestion
        let state = states[variables.ulm_card_person_entity].state;
        let option = ["home", "not_home", "unavailable", "unknown"]
        return (option.includes(state)) ? variables.ulm_translation_state : state
```

custom_components/ui_lovelace_minimalist/lovelace/translations/default.yaml
```diff
@@ -0,0 +1,104 @@
+---
+ulm_translation_engine:
+  variables:
+    ulm_translation_back: "[[[ return hass.resources[hass['language']]['ui.common.back']; ]]]"
+    ulm_translation_brightness: "[[[ return hass.resources[hass['language']]['ui.card.light.brightness']; ]]]"
+    ulm_translation_color_temperature: "[[[ return hass.resources[hass['language']]['ui.card.light.color_temperature']; ]]]"
+    ulm_translation_history: "[[[ return hass.resources[hass['language']]['ui.dialogs.more_info_control.history']; ]]]"
+    ulm_translation_hour: >
```
basbruss, line 8: Missing JavaScript template `[[[ ]]]`

custom_components/ui_lovelace_minimalist/lovelace/translations/default.yaml
```diff
@@ -0,0 +1,104 @@
...
+      var number = hass.resources[hass['language']]["ui.duration.hour"];
+      var matches = number.match(/[^{\}]+(?=})/g);
+      var hour = "Hour";
+      if (matches) {
+        var hour = matches[1];
+      }
+      return(hour);
+    ulm_translation_hours: >
```
basbruss, line 16: Missing JavaScript template `[[[ ]]]`

custom_cards/custom_card_mpse_thermostat/custom_card_mpse_thermostat.yaml
```diff
@@ -23,27 +22,17 @@ custom_card_mpse_thermostat:
...
-                label = variables.ulm_off ;
-              }else if (entity.state == 'heat'){
-                label = variables.ulm_heat ;
-              }else if (entity.state == 'cool'){
-                label = variables.ulm_cool ;
-              }else{
-                label = variables.ulm_idle;
-              }
               if(entity.attributes.temperature){
                 return (entity.attributes.current_temperature ) + '°' + ' • ' + label + ' (' + entity.attributes.hvac_action + ')';
```
basbruss, line 33: `label` is not specified anymore. Shouldn't it be `variables.ulm_translation_state`?

basbruss:
@13robin37 I have pushed a commit to this PR with additional translations for hardcoded strings. 
Could you check in you are in favor of these changes?
I have also searched the HA Lokalize for a fitting translation for `volume` but this does appear not to exist in the fronted translations.

custom_cards/custom_card_mpse_thermostat/custom_card_mpse_thermostat.yaml
13robin37: Fixed, my bad!

custom_components/ui_lovelace_minimalist/lovelace/translations/default.yaml
```diff
@@ -4,6 +4,9 @@ ulm_translation_engine:
     ulm_translation_back: "[[[ return hass.resources[hass['language']]['ui.common.back']; ]]]"
     ulm_translation_brightness: "[[[ return hass.resources[hass['language']]['ui.card.light.brightness']; ]]]"
     ulm_translation_color_temperature: "[[[ return hass.resources[hass['language']]['ui.card.light.color_temperature']; ]]]"
+    ulm_translation_status: "[[[ return hass.resources[hass['language']]['ui.dialogs.more_info_control.vacuum.status']; ]]]"
```
13robin37, line 7: Might want to switch to a general entity once available, should be fine for now.

13robin37:
@basbruss German translation is spot on!
Regarding volume: I've butchered something for German but it won't be the same for every language:
```
                      let test = hass.resources[hass['language']]['ui.card.media_player.media_volume_up'];
                      let test_ex = test.split(' ')[0];
                      console.log(test_ex);
```
Judging by https://app.lokalise.com/project/3420425759f6d6d241f598.13594006/?view=multi&search=media_volume_up there are some languages which use volume in the front and some which use it as the last word. I don't have any proper solution for the hass object right now.

basbruss:
@13robin37 Do you think it's ready to merge this amazing work?
I have made a feature request in the frontend part to make `volume` a separate translatable string with help of key-referencing. So for now we can keep it hardcoded or add it to the `ulm_language_variables`.

13robin37:
I have added a volume translation for DE and EN, should we notify others who provided the initial translation files?

basbruss:
@13robin37 I think it's okay for now. Haven't had any complaints about the hardcoded string. So we can just add it as a placeholder to other langauges.

basbruss (approved):

PR #732, opened 2022-06-06
Added/Fixed: The Light Slider can now be dragged!

AndyVRD:
Be aware that you can not use the whole slider because the padding of the thumb is set to "12px". You need to hold/tap the end of the slider to activate/use the dragging.


basbruss:
Nice work @AndyVRD!

I've tested it on android, iOS and PC/Mac. Could you also add this to the popup sliders?

basbruss (dismissed):

PR #748, opened 2022-06-16
Add custom chip card setting notification

Nik18111997:


basbruss:
Hi @Nik18111997 

Thanks for your addition!

I have a couple of remarks:

- The files should not be directly in the `custom_cards` directory but within there own folder inside the `custom_cards` directory --> (`custom_cards/custom_card_settings_notification_chip/custom_card_settings_notification_chip.yaml`)
- Please check the feedback from the Code Quality Check. The files are missing an empty last line and have trailing whitespace

PR #773, opened 2022-06-23
Add Custom Card Ristou Person

Ristou:
Hello,
Here is a new person card that might be interesting to share.
btw I was not able to run pre-commit on my local repository


custom_cards/custom_card_ristou_person/custom_card_ristou_person.yaml
```diff
@@ -0,0 +1,386 @@
...
+                [[[
+                  var location = entity.state
+                  var driving_state = "off"
+                  if (variables.ulm_custom_card_ristou_person_driving_entity !== "") {
+                    driving_state = states[variables.ulm_custom_card_ristou_person_driving_entity].state;
+                  }
+                  if (driving_state === "on" || driving_state === "true") {
+                    return variables.ulm_custom_card_ristou_person_driving;
+                  } else {
+                    if (location == "unavailable"){
```
basbruss, line 309: `ulm_unavailable` is deprecated and now part of the translation engine. Also home and not_home states are available. See example from the `card_person`:
```yaml
label: >
    [[[
      if (variables.ulm_address){
        return states[variables.ulm_address].state
      } else {
        let state = states[variables.ulm_card_person_entity].state;
        let option = ["home", "not_home", "unavailable", "unknown"]
        return (option.includes(state)) ? variables.ulm_translation_state : state
      }
```

basbruss (approved):

PR #779, opened 2022-06-23
Wrong variable names not working for custom lock card

dougmaitelli:
According to the documentation the variable names used in the template are wrong.
Example, if you try to define tap action it doesn't work since the template uses thee "ulm" prefix and the docs don't.
This PR normalizes all the variables with the "ulm" prefix.
https://ui-lovelace-minimalist.github.io/UI/usage/custom_cards/custom_card_eraycetinay_lock/

### DIFF
--- a/custom_cards/custom_card_eraycetinay_lock/README.md
+++ b/custom_cards/custom_card_eraycetinay_lock/README.md
@@ -47,9 +47,9 @@ Version: 0.0.3
   template: "custom_card_eraycetinay_lock"
   name: "Door Lock"
   variables:
-    custom_card_eraycetinay_lock_tap_control: true
-    custom_card_eraycetinay_lock_battery_level: sensor.door_battery
-    custom_card_eraycetinay_lock_door_open: binary_sensor.door_open
+    ulm_custom_card_eraycetinay_lock_tap_control: true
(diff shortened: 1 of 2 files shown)


basbruss:
Hi @dougmaitelli 

You are correct in stating that documentation asks for the prefix `ulm_`.
Changing it now will add a breaking change to already existing setups with this custom card.

dougmaitelli:
Hey @basbruss 

Yeah, I know, but this change is not just fixing the pattern, it is also fixing some variables that are not currently working as expected.

I think this could be merged for a next minor release with a "breaking changes" note in the release notes.

basbruss:
@dougmaitelli Which variables are not working as expected?
I personally don't use the card, but judging by the source code it should be fine when used with the accompanying language files.

dougmaitelli:
@basbruss the tap action (and there is one other) are currently using the "ULM" prefix in the code but that is not what is documented.

basbruss:
@dougmaitelli 

I agree: that documentation is not in line with variables in the code. And therefore better to convert everything to the prescribed method. 
I will take note of the breaking change.
Could you re-branch it to the `release` branch? So the changes are only applied when the release notes with the breaking change are out.

PR #823, opened 2022-07-01
Updated Minimal Door Lock to V2

Gilganik:
Updated Minimal Door Lock to V2


basbruss (approved):

Gilganik:
Hey @basbruss I see from the link: https://ui-lovelace-minimalist.github.io/UI/usage/custom_cards/custom_card_nik_door/#description that the page is not updated with version 2.0. What can I do? Can you check it?

PR #828, opened 2022-07-02
New more featureful custom_card_esh_room version (2.0.0)

mpeterson:
This change adds support for a switch to control climate and also to remove the need to have an entity associated. Also it makes both buttons optional for more flexibility.

This is a breaking change.


mpeterson:
@EverythingSmartHome I hope you are okey with the changes I introduced to the card you contributed

GuyLewin:
@mpeterson this seems super useful! I'll definitely use it when it's merged. Thanks!!

basbruss (approved):

PR #853, opened 2022-07-10
Improve card_person card aesthetics

Petapton:
- Use zone friendly name in label when state is a real zone
- Match zone1 and zone2 by checking if person is in zone.attributes.persons


basbruss:
@Petapton 

PR's for official cards should be directed to the `release` branch as stated in the `contribution.md`

Petapton:
Sorry, my bad.

PR #889, opened 2022-07-25
Fix typo in card_battery.md

Michelli:
Found/fixed a typo in one of the card pages


basbruss:
Hi 

Hi,

Changes to `/docs/usage/custom_cards` will not hold and are overridden by the readme files in the `/custom_cards` directory. So to change the custom_cards docs you should change the corresponding readme files in the `/custom_cards/` directory.

Issue #886, opened 2022-07-25
ulm_name_tap_action not working on card_binary_sensor


bokub:
**Describe the bug**
The "name" tap action cannot be customized on "binary sensor" card.

**To Reproduce**

1. Create a new [input boolean](https://www.home-assistant.io/integrations/input_boolean/)
2. Create a new card with the following YAML code:
```yaml
    - type: "custom:button-card"
      template: card_binary_sensor
      variables:
        ulm_card_tap_action: more-info
        ulm_name_tap_action: more-info
        ulm_icon_tap_action: more-info
      entity: input_boolean.my_new_input_boolean
```

4. Click on the name of the input in the new card

**Expected behavior**
It should open the "more-info" popup. Instead it toggles the input.

**Additional context**

**N.B**: The `ulm_card_tap_action` and the `ulm_icon_tap_action` work well, only the `ulm_name_tap_action` is broken.

Issue #884, opened 2022-07-24
Custom Actions overwrite on single card won't work for name and icon.


matrixx567:
**Describe the bug**
The specific overwrite for the custom actions on a card won't work on the name and the icon fields. For the card the custom overwrite of the action works as expected.

**To Reproduce**

Configuration to open a popup on tap action for a sprecific card
```yaml
      - type: 'custom:button-card'
        template: 
          - card_generic
        entity: sensor.tempsensor_office_1_temperature
        variables:
          ulm_card_generic_name: "Temperatur"
          ulm_card_tap_action: "adaptive"
          ulm_name_tap_action: "adaptive"
```

**Expected behavior**
The popup should be opened if I tap on the name

**Additional context**
I think the variable overwrites must be redirected  to the custom fields in the `icon_more_info_new.yaml` and in the 'card_generic.yaml`.

PR #878, opened 2022-07-22
Custom popups for cards that uses custom actions.

matrixx567:
Hy,

I really like the new custom actions feature but currently I miss the possibility using custom popup cards in an easy way.
This pull request implements the possibility to use custom popups with cards that already implements the custom actions feature. 

A popup for a card can be simply configured using variables. See the example below:

```yaml
    - type: 'custom:button-card'
      template: card_power_outlet
      entity: switch.power_outlet_livingroom
      variables:
        ulm_card_power_outlet_consumption_sensor: sensor.power_outlet_livingroom_consumption
        ulm_card_power_outlet_name: Power Outlet Livingroom
        ulm_custom_popup:
            template: "popup_power_outlet_stats"
            popup_variables:
              ulm_popup_power_outlet_sensor1: sensor.power_outlet_livingroom
              ulm_popup_power_outlet_sensor2: sensor.power_outlet_livingroom_consumption
              ulm_popup_power_outlet_graph_sensor: sensor.power_outlet_livingroom
              ulm_popup_power_outlet_entity: switch.power_outlet_livingroom        
```

For using a custom popup the variable `ulm_custom_popup` must be defined. This variable is a dictionary containing the name of the popup's custom button template (`template`). It is also possible to inject the necessary variables for the popup using `popup_variables`.

What do you think about this feature? 

(I'm not the JavaScript expert, so I would be open for suggestions and improvements.)

PR #893, opened 2022-07-26
Fix custom_action overwrite

schumijo:
Solve #884 and #886

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_battery.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_battery.yaml
@@ -25,6 +25,9 @@ card_battery:
             card:
               type: "custom:button-card"
               variables:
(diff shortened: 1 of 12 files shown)


bokub:
I confirm it solves #886 :tada:

schumijo:
navigate_path are missing
Will add it later....

matrixx567:
Hi @schumijo,

I think the variables can be set in the `icon_more_info_new` and the `icon_more_info_alert` for the custom fields `item1` and `item2`. The template will already set these variables. So there will be no need to change every single card.

schumijo:
Hi @matrixx567 
I already tried this yesterday without success. Is it working on your side ?

matrixx567:
Hi @schumijo ,

I've implemented a working version in my PR #878.
The idea is quite simple. The icon_more_info_new implements the `ulm_custom_actions` template and redirects it to the custom fields.
I've also implement the custom actions on the graph card which uses a generic card as custom field.

Currently I have tested my implementation with the generic_card, graph_card and the power_outlet_card.

Take a look.

schumijo:
Refer to #878

PR #916, opened 2022-07-30
Improve zones

rob1998:
Eliminate the need for the ulm_card_imswel_person_zones variable, by automatically querying all zones from ha states


basbruss:
Hi @rob1998 

Could you also change the documentation? As this is a change to existing setups.

rob1998:
True, completely forgot about it. Something like this?

basbruss (approved):

PR #948, opened 2022-08-17
Creation of popup thermostat which uses preset_mode instead of state.

MatthewSean:
I created this for my own use, and thought I should share.  Hopefully, I have done this in the correct way.  If not let me know.


basbruss:
@schumijo What do you think of this addition? Can we combine it with the original one?

basbruss:
> @basbruss
> 
> Yes seems to be possible to combine with original one by adding a variable to choose between hvac mode and preset mode.

Okay I will try to add this feature to the original one with some commits to this PR. @MatthewSean are you okay with that?

MatthewSean:
Please feel free to add, sorry for the late reply.

basbruss:
@MatthewSean I think I am done with combining both features. Could you test if you are satisfied with the result?

To switch to preset mode just add to your card config
```yaml
variables:
  ulm_card_thermostat_enable_popup: true
  ulm_card_thermostat_preset_mode: true
```

MatthewSean:
Please go ahead.

basbruss (approved):

Issue #978, opened 2022-09-06
Light Popup not correctly aligned on mobile


matrixx567:
**Describe the bug**
On a mobile display the light popup isn't correctly aligned if switched between brightness and color.

**Screenshots**
![image](https://user-images.githubusercontent.com/4608946/188623822-b49d0472-8739-4a42-a50c-c755195639ee.png)
![image](https://user-images.githubusercontent.com/4608946/188623876-82cc6df9-05f0-42df-9df5-196ecc9e118f.png)

@schumijo, @AndyVRD

matrixx567:
Also on a large display the left and the right margins are not the same.

![image](https://user-images.githubusercontent.com/4608946/188624571-11d2c536-6ff7-40a9-b86d-f69a8746f535.png)

basbruss:
Hi @matrixx567 

It also happens to me on the iOS app. But either in my dev-env or personal instance in chrome, it displays correctly when simulating a smartphone screen.
<img width="330" alt="image" src="https://user-images.githubusercontent.com/68892092/188626956-a9e2045e-e8cb-4288-8f4e-dabd2e02a347.png">

AndyVRD:
@basbruss @matrixx567 I already upgraded all the sliders to v2 and this will fix the issue I think. I’m not sure because I don’t have a light with only brightness and temperature available so I can’t check it. I will finally release these updates tomorrow and I will ask Bas to check if it fixes the alignments.

currydani:
For me the Popups also got some problems with the layout on desktop, the size of the color wheel for the light popup and the rounded corners are not displayed correctly.

![image](https://user-images.githubusercontent.com/78862555/188695318-41d46ed5-f37d-4c66-8346-aeceb9a0f00b.png) ![image](https://user-images.githubusercontent.com/78862555/188695357-ea125f27-44be-4faf-8d31-d5a8739e7466.png)

basbruss:
@AndyVRD @matrixx567 I think I found the cause of the shifting alignment to the right.
<img width="1002" alt="image" src="https://user-images.githubusercontent.com/68892092/188807324-795d39a0-7c5e-487b-967f-8da58b617cae.png">

I played a bit with the container width of the `custom:button-card`-popup. The full width is not acknowledged by the browser_mod div. 
So we need to find a way to force browser mod to use the width set in `custom:button-card`-popup instead of it's own generated value.

matrixx567:
@basbruss 

[...]

PR #979, opened 2022-09-07
Update browser mod service

basbruss:
- fix max width of 600px
- fix no full screen on mobile
- fix light popup on mobile
- Update slider in popup to V2


basbruss:
@matrixx567 Can you check if this PR solves all the issues reported in #978

schumijo:
Hi @basbruss 
I am in holidays so I can't do any reviews/tests

PR #980, opened 2022-09-07
Binary sensor alert card state

csuich2:
This introduces a variable that specifies which state icon_alert and icon_info_alert should alert on. This is useful when you want to alert on a state other than "on". For example, I use this to alert when my automation is "off".

Need to setup a test environment, but this is a handy change that I've had in a local custom card. I figured it made more sense to put into the base cards than a custom card.

If you're open to this change then I can setup the dev environment for a full test.

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/internal_templates/icon_alert.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/internal_templates/icon_alert.yaml
@@ -7,7 +7,7 @@ icon_alert:
   show_name: false
   state:
     - operator: "template"
-      value: "[[[ if (entity.state == 'unavailable' || entity.state == 'on'){ return true; } ]]]"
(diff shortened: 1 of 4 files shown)


basbruss:
Hi @csuich2 

Thank you for improving this card.

Binary Sensors can only have three states: `on`; `off`; `unavailable`. Therefore to make it user-friendly and less error-prone you can consider changing the added variable to a boolean variable that can revert checking for the `on` state to the `off` state by setting `true` as the value for the variable.

Example:
```yaml
value: > 
  [[[ 
    var check_state = 'on'
    if(variables.ulm_icon_alert_revert_state){
     check_state = 'off';
    }
    if (entity.state == 'unavailable' || entity.state == check_state){ return true; } 
  ]]]

csuich2:
@basbruss Great point! I've updated with a boolean flag and verified the behavior with a devcontainer.

basbruss:
@csuich2 Works perfect!

It will be available with the next release!

basbruss (approved):

Issue #983, opened 2022-09-09
Tap action does nothing on name


damusmf:
**Describe the bug**
updated minimalist to 1.1.3 and browser mod to 2.1.3 and now tap action does not work on any cards in the name area. i can verify the other actions work as i get the haptics and pop up cards where configured, as well as toggle working on the icon and card.  But toggle does not seem to function on the name.  I have cleared cache and tested on multiple browsers and the HA App on android.

**To Reproduce**

update to latest version of everything, tap on name field of light card.

**Expected behavior**
expect it to toggle the device or do the action based on custom_actions.yaml

**Screenshots**

**Additional context**
```
  - type: grid
    columns: 2
    square: false
    cards:
      - type: "custom:button-card"
        template: card_light
        entity: light.office
        variables:
          ulm_card_light_name: Ceiling
      - type: "custom:button-card"
        template: card_light
        entity: light.office_lamps
        variables:
          ulm_card_light_enable_popup: true
          ulm_card_light_name: Lamps
      - type: "custom:button-card"
        template: card_light
        entity: light.office_closet
        variables:
          ulm_card_light_name: Closet
```

custom_actions.yaml
```
---
ulm_custom_actions:
  variables:
    ulm_card_tap_action: "toggle"
    ulm_card_hold_action: "popup"
    ulm_card_double_tap_action: "adaptive"
    ulm_icon_tap_action: "toggle"
    ulm_icon_hold_action: "more-info"
    ulm_icon_double_tap_action: "adaptive"
    ulm_name_tap_action: "toggle"
    ulm_name_hold_action: "popup"
    ulm_name_double_tap_action: "adaptive"
    ulm_card_tap_haptic: "light"
    ulm_card_hold_haptic: "light"
    ulm_card_double_tap_haptic: "none"
    ulm_icon_tap_haptic: "light"
    ulm_icon_hold_haptic: "light"
    ulm_icon_double_tap_haptic: "none"
    ulm_name_tap_haptic: "light"
    ulm_name_hold_haptic: "light"
    ulm_name_double_tap_haptic: "none"
```

thibbocli:
I have the same bug.  Here is a bit more informations : 
the hold action and the double tap works fine for me.
If it can help, here is the error in the console : 

`
button-card.js?hacstag=146194325342:425 Uncaught ButtonCardJSTemplateError: SyntaxError: Unexpected token 'return' in 'var action = variables.ulm_name_tap_action;

[...]

PR #988, opened 2022-09-11
Fix syntax error for name tap action

lkmgr:
There is one `}` bracket too many in the `actions_name.yaml` file - removing it should fix the error described in #983


jklimek (approved):
This change indeed fixes tap_action bug mentioned in https://github.com/UI-Lovelace-Minimalist/UI/issues/983. I tested it in my local environment with positive results.

jklimek (approved):

basbruss (approved):

PR #1032, opened 2022-09-29
[popup_mediaplayer]Add auto-generated source selector

basbruss:
Still TODO:

- [x] add source selector to `card_media_player`
- [x] change `play/pause` behavior to `start/stop` when using a url-stream (known issue with at least HEOS-devices)

https://user-images.githubusercontent.com/68892092/193017200-282e348d-bc0f-4127-b595-1ba0f2a04dee.mov

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_media_player.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_media_player.yaml
@@ -437,6 +437,24 @@ card_media_player:
                   data:
                     large: true
                     hide_header: true
+                    style: |
+                      --popup-background-color: var(--primary-background-color);
+                      --popup-border-radius: 20px;
+                      --popup-padding-x: 24px;
+                      --popup-padding-y: 20px;
+                      --popup-max-width: auto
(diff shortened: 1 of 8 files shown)


schumijo:
Hi @basbruss 

I did some tests :
- Popup is broken if media_player entity doesn't have a source_list attribute
- Some trouble with displayon desktop only (seems good on mobile) : 
Capture d’écran 2022-10-11 111227

basbruss:
Thanks @schumijo 
> * Popup is broken if media_player entity doesn't have a source_list attribute
I can make everything optional and make the same solution as in the media_play_card itself where it shows the more-info-dialog if there is no source list attribute.

> * Some trouble with display on desktop only (seems good on mobile)
That is probably related to the `absolute` position to the bottom of the screen and the relative height of the card to screen-size. So I will tweak the position to be absolute using the dimensions of the popup itself instead of the screen.

basbruss:
@schumijo 
I have pushed some commits to fix the issues you have found.



> Why don't you use viewport height as I did in other popups ?

I have now and set the content-grid-container to be able to scroll instead of the whole popup-card. This removes the need for a fixed footer position.

schumijo:
Hi @basbruss 
Everything seems OK now

schumijo (approved):

PR #1046, opened 2022-10-11
[card_person] Exclude passive zones

freddanastrom:
Excluding passive zones and making the persons attribute in zones optional, since it breaks the card if the attribute for some reason is missing.


basbruss:
@freddanastrom Just wondering. Is it also fixed without filtering the `passive` attribute? But only with adding the additional optional chain to the condition?

freddanastrom:
@basbruss Yes, the optional condition to persons attribute is actually enough.
But I also thought the exclusion of passive zones was a good idea to implement, since HA documentation says passive zones are hidden from the frontend and not used for device tracker names.

PR #1048, opened 2022-10-13
Fixed depracated use of speedtestdotnet.speedtest service and added dutch language

delcara:
HA notifies:

The speedtest service is being removed
Update any automations or scripts that use this service to instead use the homeassistant.update_entity service with a target Speedtest entity_id. Then, click SUBMIT below to mark this issue as resolved.

### DIFF
--- a/custom_cards/custom_card_irmajavi_speedtest/custom_card_irmajavi_speedtest.yaml
+++ b/custom_cards/custom_card_irmajavi_speedtest/custom_card_irmajavi_speedtest.yaml
@@ -73,7 +73,9 @@ custom_card_irmajavi_speedtest:
         type: "custom:button-card"
         tap_action:
           action: "call-service"
-          service: "speedtestdotnet.speedtest"
+          service: "homeassistant.update_entity"
+          service_data:
+            entity_id: sensor.speedtest_download
         color: "var(--google-grey)"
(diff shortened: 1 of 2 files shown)


custom_cards/custom_card_irmajavi_speedtest/custom_card_irmajavi_speedtest.yaml
```diff
@@ -73,7 +73,9 @@ custom_card_irmajavi_speedtest:
         type: "custom:button-card"
         tap_action:
           action: "call-service"
-          service: "speedtestdotnet.speedtest"
+          service: "homeassistant.update_entity"
+          service_data:
+            entity_id: "sensor.speedtest_download"
```
basbruss, line 78: Hardcoding an entity is not advised. The sensor can be different for each user. 
Also, this is a breaking change to use a sensor instead of the integration and should be reflected in the accompanying documentation.

delcara: I changed the hardcoded sensor to use the variable and modified the readme with the breaking change that the speedtestdotnet.speedtest service is nog longer used.

basbruss (approved):

PR #1081, opened 2022-11-06
Update eraycetinay custom lock card to include garage door cover states

thinkjk:
Currently for a cover the states are closed and opened. Adding this to the card in order to have colors set properly. Original code by @sisimomo 

Also updated 2 files to fixing missing returns which was causing one of the checks to fail

image


thinkjk:
The failed check looks to be an issue with the check code itself rather than my PR

basbruss (approved):

Issue #1094, opened 2022-11-18
custom_card_esh_welcome displays the wrong language when there is more then one translation file present in the Minimalist folder


Marck:
The custom_card_esh_welcome displays only morning, evening, afternoon greetings from the TR language yaml (I think it's turkish) and doesn't change while it should change.

Is there a missing settings or something? I could not find anything in the docs

Marck:
For reference
<img width="509" alt="image" src="https://user-images.githubusercontent.com/18088281/202790150-554d150f-a998-4088-9386-6835d2fb6b83.png">

xm4rcell0x:
Have you changed your language settings in "Integrations"? 
There's an option to select your language. 
![image](https://user-images.githubusercontent.com/17478462/206937393-38638422-736c-430e-81d2-7877caf344de.jpeg)

Marck:
Yes, It's at GB but that doesn't change a thing sadly
<img width="389" alt="image" src="https://user-images.githubusercontent.com/18088281/207016847-8610790e-defb-4798-9fb0-8518228824b3.png">

Marck:
Closing this for I have a workaround

PR #1106, opened 2022-11-29
Added language field for use in custom_cards

Marck:
I added for each language a language field describing the language name in the language it's in.

*Why is this usefull?*
I found that custom_cards use the `ulm_language_variables` variable, but if there is more then one translation file present, Minimalist grabs the last file loaded in (in my case Turkish). This is a fix that enables bugs like #1094 to be fixed easier in my opinion.

So adding this *should* enable the opportunity to do logical checks like `if ulm_language == hass_language_variable` to determine the correct language.

This PR is blocked by #1098 or needs that PR pulled in first before merging. Let me know what the accepted approach is (don't want to steal the credits from @ulug79)

Note: please review the language names as my mother language is Dutch and I used Google Translate to translate the names

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/translations/cn.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/translations/cn.yaml
@@ -1,6 +1,7 @@
 ---
 ulm_language_variables:
   variables:
+    ulm_language: "中國人"
     ulm_currency: "￥"
     ulm_updates_available: "有更新!"
     ulm_no_updates_available: "无更新"
--- a/custom_components/ui_lovelace_minimalist/lovelace/translations/cn.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/translations/cn.yaml
@@ -1,6 +1,7 @@
 ---
 ulm_language_variables:
   variables:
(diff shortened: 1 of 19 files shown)


Marck:
Fyi; this is how it looks if you display it in a card (nothing special, not the intended use of this PR but maybe someone wanted to see how it looks like):
image

ulug79:
> I added for each language a language field describing the language name in the language it's in.
> 
> _Why is this usefull?_ I found that custom_cards use the `ulm_language_variables` variable, but if there is more then one translation file present, Minimalist grabs the last file loaded in (in my case Turkish). This is a fix that enables bugs like #1094 to be fixed easier in my opinion.
> 
> So adding this _should_ enable the opportunity to do logical checks like `if ulm_language == hass_language_variable` to determine the correct language.
> 
> This PR is blocked by #1098 or needs that PR pulled in first before merging. Let me know what the accepted approach is (don't want to steal the credits from @ulug79)
> 
> Note: please review the language names as my mother language is Dutch and I used Google Translate to translate the names

Hi, don't worry about my credits. All PR from me and others only for community! :)

Marck:
This will never get merged I'm afraid so closed.

PR #783, opened 2022-06-24
Custom name (issue #778)

schumijo:
PR to solve issue #778 

Done :
- card_battery : Variable `ulm_card_battery_name` added

PR #1107, opened 2022-11-30
custom_card_paddy_waste_collection- fixed broken variables to point to entity

jameshirka:
The custom card was broken as per:
https://github.com/UI-Lovelace-Minimalist/UI/pull/783

Fixed this by setting the variable to the entity.


AGatGitHub:
Works for me, except that the notification exclamation mark is behind the icon instead of in front of it (see: [https://community.home-assistant.io/t/lovelace-ui-minimalist/322687/3067](url)). Did you observe this behaviour as well?

basbruss (approved):

PR #134, opened 2021-12-22
All in one Fan Card

sildehoop:
I have made a Fan Card and Fan Slider Card with temperature and humidity label.
card_fan
card_fan_slider

To do:
- [ ] Merge fan slider
- [ ] Merge fan slider horizontal
- [ ] Merge fan collapse
- [ ] Merge fan oscillate

Issue #107, opened 2021-12-19
Adjust Custom Fan Card for Air Purifiers


13robin37:
Currently the custom_card_saxel_fan by @saxel works only for traditional fans (which is fine), however Air Purifiers like the [Xiaomi](https://www.home-assistant.io/integrations/xiaomi_miio/#xiaomi-air-purifier-air-humidifier-and-standing-fan) ones work with the fan integration for the most part as well.

In my opinion the current button which triggers call-service fan.osciallate should switch between fan modes Off and Auto for purifiers. Should I set up a new custom card or integrate something within the existing card?

CM000n:
Thank you so much for figuring this out! In my opinion, we should avoid redundant cards that look the same and do the same thing, just a little differently.
So I would rather prefer the second solution by feeling. But let's wait for the feedback from @saxel. Maybe he has another opinion about it.

saxel:
Hello, sorry for the delayed response (just got back from a short vacation, also dealing with aftermath from flooding reaching into my apartment). I also have the same brand of air purifier but I haven't found a use case for controlling fan speed and off/auto. Mine is strictly controlled via automations.

I suppose if you want to control the air purifier fan speed with a slider then I agree with @CM000n that the 2nd suggestion makes sense, if it is okay that the slider is visible even when the air purifier fan speed is set to Auto.

@13robin37 do you want to give it a try with implementing? Otherwise I can have a look either later today or within a couple of days depending on my time available.

13robin37:
@saxel I will try to implement something in the next few days when I have some free time.

13robin37:
Apologies for the delay on this one, we've moved and I'm slowly picking up speed on Home Assistant projects again! 🙃

lucasfr:
I'm not sure whether I am missing something but it would also be good to hide the oscillate button. Could not find anything in the docs.

wilbiev:
Cleaning up issue list

PR #1116, opened 2022-12-07
Saxel Fan Updates

benbur98:
I have worked on a few updates to @saxel fan card.
Have incorporated some of @sildehoop work with the percentage/humidity labels (these can be optioned out if not desired), from his PR #134 

I have added the option to remove the Oscillation button as requested in #107 . With changing the Service of this button it is more complicated, could someone with an air purifier comment on what service (and it's data) they would specifically like - I have started to add this option but currently only oscillation will work (as previous) as i haven't added anything to change the Service data.

Planning on implementing the Horizontal card layout also.

### DIFF
--- a/custom_cards/custom_card_saxel_fan/README.md
+++ b/custom_cards/custom_card_saxel_fan/README.md
@@ -109,6 +109,24 @@ This card needs the following to function correctly:
     <td>true/false if the fan speed row should collapse when the fan is turned off. Default: true</td>
   </tr>
   <tr>
+    <td>ulm_show_button</td>
+    <td>false</td>
+    <td>no</td>
+    <td>Set to false to remove the Button and have a full width Slider. Default: true</td>
+  </tr>
+  <tr>
+    <td>ulm_button_icon</td>
+    <td>mdi:button</td>
+    <td>no</td>
+    <td>Set the icon for the Button. Default: mdi:rotate-3d-variant</td>
+  </tr>
+  <tr>
+    <td>ulm_button_service</td>
+    <td>fan.toggle</td>
+    <td>no</td>
+    <td>Name ofthe Service Call on button press. Default: fan.oscillate</td>
+  </tr>
+  <tr>
     <td>oscillate_attribute</td>
     <td>oscillate</td>
     <td>no</td>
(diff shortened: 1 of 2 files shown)


benbur98:
Quick and dirty Horizontal Mode added.
As mentioned in #134 , I think the card needs rebasing off of the Light card as it is significantly different in what is fairly similar operation. Will look into this.

shift-del1:
@benbur98: I have Xiaomi Mi Air Purifier 2S and 3, so I can do the tests. From my point of view the only reasonable servise to be assigned to the button would be the looping between presets, this is what the button/touch screen does too. One press, Auto, press again, Silent, press again Manual/Favorite and so on (or if there are levels, those ones). In my case the attributes are the following:
  preset_modes:
    - Auto
    - Silent
    - Favorite
    - Idle/Fan (2S/3)
  preset_mode: Favorite
so fan.set_preset_mode would seem to be the right choice (personally I don't modify the preset manually). Favorite and Fan modes have number entities, so long pressing the button could be used to alter somehow the number entity assigned to that specific mode (through number.set_value). These are just ideas, persionally I would remove the button completely 😊

benbur98:
Thanks for the info. Implementing auto mode toggle should be doable. The trouble really is keeping it consistent between different fans - unfortunately there is no ‘get available presets’ option so without restricting it by variables it’s a very specific option.
Also I have added the option to remove the button now if you like.

shift-del1:
@benbur98 I replaced the yaml with yours, but I got an error for label (I defined the variables for temp and hum attributes and the one to remove the button. After hardcoded a value for label, I got this:
TypeError: Cannot assign to read only property '0' of object '[object Array]'
Any suggestion where did I go wrong? Thank you.

benbur98:
@shift-del1 Thanks for checking this out. How did you define the variables, can you share your YAML? Also what do you mean by 'hardcoded a value for label'.

benbur98:
Note: This check is only spelling due to the spelling of 'Collapsable' in the Doc - I Have spelt it this way as changing the variable spelling to be correct 'Collapsible' is not worth the breaking change.

shift-del1:
@benbur98: I reverted back my config to the original fan card, so I can not show my YAML (sorry). "Hardcoded" means I replaced the label section with a string (so skipped the whole function which defines the string to show as a label). Technically it should simplyfy the card while loosing its flexibility, but for my test it should have been enough (but it wasn't :-D). I'm sure it is hard to help without data, just wanted to ask if you faced the same previously.

Issue #1042, opened 2022-10-10
Volume buttons in card_media_player (and other improvements)


antonio1475:
**Is your enhancement request related to a problem? Please describe.**
The Media_player card has some shortcomings for me:
1. Volume slider pointless when needing to move in small increments, and always moving between 20-40% volume at max.
2. Controls & text not readable with many album covers (different colors, letters, etc).

**Describe the solution you'd like**
Update to the existing card. I'm sorry that I don't understand how contibuting in GitHub works...

**Describe alternatives you've considered**
I created a custom card to add:
1. Volume buttons (1% for TVs, 5% for speakers)
2. Mute acts as toggle (check line 450)
3. background semi-transparent layer for readability (the css already had 0.15 opacity but wasn't displayed. Added `background-blend-mode: "multiply"` to make it effective)
4. Adjust Pause button to use `mdi:play-pause` by default, and then added status playing to show `mdi:pause` (eg. a TV that's not reporting the playback status was always showing mdi:pause, even when playing).
5. Added/changed some hold_actions to toggle the player.

See yaml comparison here: https://www.textcompare.org/yaml/?id=637f742eb35ae78d291d9cff

@rairulyle or anyone if you want to use this for an appropiate code change, please by all means. 
Also, the mute icon could probably change based on the mute status, but I don't know how to pull an attribute for the status.

Note: I made some changes just for me, that might not want on the public version (eg changed some variable defaults, aspect ratio, hold to toggle, etc)

**Additional context**
Now vs proposed changes:
![image](https://user-images.githubusercontent.com/63130724/194899774-2761c924-a634-4860-ae77-4c7ee744128c.png)
_Note: screenshot made with opacity 0.25 for background, plus another 0.25 for the buttons. yaml above has 0.15 & 0.2 respectively._

antonio1475:
I just saw by chance that background-blend-mode: justify was removed here: https://github.com/UI-Lovelace-Minimalist/UI/issues/970#issuecomment-1242976950

antonio1475:
I made further changes so that the volume changes by 1 for TVs, and 5% for speakers.
```
                  volume_level: |
                      [[[
                        var volume = states[entity.entity_id].attributes.volume_level;
                        if (states[entity.entity_id].attributes.device_class === "tv"){
                          volume = states[entity.entity_id].attributes.volume_level + 0.01;
                        }
                        if (states[entity.entity_id].attributes.device_class === "speaker"){
[...]

PR #1117, opened 2022-12-07
Media Player Card Volume Buttons

benbur98:
Added a new row to use Volume Buttons (Mute, Up and Down) to the Card.
Can be as well as or instead of the Slider, just an additional row.

Majority of this work is taken from #1042 so big thanks to @antonio1475 for that. I have not implemented all of his updates a I've tried to keep the card as similar as possible to the previous version.
Included:
- Volume Buttons
- Toggle Mute
- Volume Adjust Amount (1% for TV, 5% for Speaker, 2.5% Receiver or Default)


basbruss:
part of #1114

PR #1118, opened 2022-12-07
custom_bar_card Icon Choice

benbur98:
Simple fix to allow icon to be overwritten for custom_bar_card

(Apologies for numerous PRs... have kept separate for each card but might not be the preferred option)


basbruss:
part of #1114

PR #1123, opened 2022-12-21
Improve card_scene

aehru:
Combine custom card_scenes and card_scenes_welcome mainly for color possibilities.

I deleted the original custom card_scenes but feel free to revert this.
Maybe we can also delete the card_scenes_welcome in order to have only one of those cards ?


basbruss:
This will break `card_welcome_scenes`

aehru:
I'm confused on what will break `card_welcome_scenes`.
I've done some tests and it seems fully compatible for my use, could you help me find what's wrong ?

I haven't pushed the modification of `card_welcome_scenes` waiting for other feedback.

basbruss:
I was confused by this I think and didn't noticed the file was not really deleted with a commit. 
> Maybe we can also delete the card_scenes_welcome in order to have only one of those cards ?
Basically with this PR there are 2 very similar cards in the base repo.
Deleting `card_scenes_welcome` will break `card_welcome_scenes`.

I think the way forward is to keep your `card_scenes` modification and make `card_scenes_welcome` backward-compatible with `card_scenes` by changing `card_scenes_welcome.yaml` to;

```yaml
card_scenes_welcome:
  template: "card_scenes"
```
This will allow this PR to be useful without breaking any config of users

wilbiev:

PR #1144, opened 2023-01-08
graph_card: use variable to adjust line_width

shogun160:
use variable ulm_card_graph_line_width to adjust line width of the graphs


basbruss:
Duplicate of #1124

PR #1151, opened 2023-01-15
increase number of colors of the navigation button

auadix:
With this change you can chose from all those available colors (red, pink, purple, deep-purple, indigo, blue, light-blue, cyan, teal, green, light-green, lime, yellow, amber, orange, deep-orange, brown, grey, blue-grey, black, white) instead of being limited to just 6 (blue, red, green, yellow, pink, purple).

### DIFF
--- a/custom_cards/custom_card_esh_welcome/custom_card_esh_welcome.yaml
+++ b/custom_cards/custom_card_esh_welcome/custom_card_esh_welcome.yaml
@@ -6,35 +6,35 @@ card_esh_welcome:
     entity_1:
       color: >
         [[[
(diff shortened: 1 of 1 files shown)


auadix:
Other advantage of this change, the navigation button in the Home Assistant app now show the colors correctly. (tested on Android 13 and iOS 16.2)

basbruss:
Will break configs that have not set a color_value due to absent of a default

ABUAIJUTIE:
Definitely needed! 
It is so weird to limit us in these 6 colors. I mean not only the navigation button, but also the whole UI Minimalist...

wilbiev:

PR #1152, opened 2023-01-15
Added Group Entity Support

stefanschlipfi:
Added Group Support for Custom Card: custom_card_yagrasdemonde_lights_count

### DIFF
--- a/docs/usage/custom_cards/custom_card_yagrasdemonde_lights_count.md
+++ b/docs/usage/custom_cards/custom_card_yagrasdemonde_lights_count.md
@@ -56,6 +56,20 @@ template:
           {{ covers | selectattr('state','eq','open') | list | count }}
 ```
 
+### Using Groups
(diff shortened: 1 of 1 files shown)


docs/usage/custom_cards/custom_card_yagrasdemonde_lights_count.md
```diff
@@ -56,6 +56,20 @@ template:
...
+If you want a group entity instead of a single explicit.
+
+```yaml
+template:
+  - sensor:
+      - name: "All Lights on"
+        unique_id: sensor_all_lights_on
+        icon: mdi:lightbulb-on-outline
+        state: >
+          {{ expand('group.all_lights') |  selectattr('state','eq','on') | list | count }
```
basbruss, line 70: Maybe better to give an example with the `light` groups instead of the general group

PR #1177, opened 2023-02-15
Fix: Mobile Popups

basbruss:


albinmedoc (commented):
I don't know if it's possible. But there should be a generic popup that all others inherit from.
That would result in not making these types of changes in every popup.

PR #1194, opened 2023-03-03
Card Fan as a New Base Card

benbur98:
Fix Issue #1191 

I have created a new card based largely from @saxel fan card. This has been completely re-based to follow the modern format from the Light Card. I feel this card is now ready to be a base card and not a custom-card.

I believe this should Close and Replace PR #134.

PR #1204, opened 2023-03-08
Card Cover Custom Colors

benbur98:
Cover Updates:
 - Ability to set any color instead of just fixing it to Blue
 - Slider Override Min/Max Values

CSS Grid Code Updated slightly on Both Cards to be more flexible for future additions of rows

- [X] The code change is tested and works locally.
- [X] There is no commented out code in this PR.
- [x] I have followed the [contribution guidelines](https://github.com/UI-Lovelace-Minimalist/UI/blob/main/.github/CONTRIBUTING.md)

Please feel free to test these changes out :)

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_cover.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_cover.yaml
@@ -6,8 +6,8 @@ card_cover:
     - "ulm_translation_engine"
   variables:
     ulm_card_cover_name: "[[[ return entity.attributes.friendly_name ]]]"
-    ulm_card_cover_icon: false # "[[[ return entity.attributes.icon ]]]"
-    ulm_card_cover_color: false
+    ulm_card_cover_icon: false
+    ulm_card_cover_color: "blue"
     ulm_card_invert_percent: false
     ulm_card_cover_display_left_right: false
     ulm_card_cover_garage_large: false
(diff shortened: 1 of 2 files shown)


basbruss:
@benbur98 Thank you very much for your contributions.

Please create different PR's for each different card. This makes everything trackable. Also, you seem to have ignored the contribution guidelines with selecting the correct branch:
 

> - [ ] This PR is for a custom-card or documentation change and therefore directed to the main branch.
> - [ ] This PR is for a official card or any other directly to the integration related change and therefore directed to the release branch.

Same holds for #1194

benbur98:
@basbruss Sorry, you're quite right... I'll re-read the guidelines now to get it right from now on!

custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_cover.yaml
```diff
@@ -9,6 +9,7 @@ card_cover:
     ulm_card_cover_icon: false
     ulm_card_cover_color: "blue"
     ulm_card_invert_percent: false
```
benbur98, line 11: Not sure your thoughts on breaking changes, but this variable has only been in since v1.2.0 and my PR #1114 - so it's unlikely it's been heavily used yet. Could just include the correct 'ulm_card_cover_invert_percent'? This would clean up some of the logic where we have to check twice below...

benbur98 (commented):
One small query. Otherwise I've tested and this works well for me. Thanks

custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_cover.yaml
basbruss: We cannot be sure how many users have used the variable.
To maintain a streamlined update it's better, in my opinion, to add some legacy support.
At least it will reduce (simple) support questions and issues.

benbur98: Yeah okay agreed :)

basbruss (approved):

PR #1209, opened 2023-03-10
Card_light Rework color background

basbruss:
Make background colors more inline with examples found on behance

Example with `ulm_card_light_force_background_color: true`
image

Old behavior:
image

- adds variable `ulm_card_light_color` to set manually the color when a light/switch is on

## Additional information

- This PR fixes or closes issue: fixes #
- This PR is related to issue:
- Link to documentation pull request:

## Checklist

- [x] The code change is tested and works locally.
- [x] There is no commented out code in this PR.
- [x] I have followed the [contribution guidelines](https://github.com/UI-Lovelace-Minimalist/UI/blob/main/.github/CONTRIBUTING.md)
    - [ ] This PR is for a custom-card or documentation change and therefore directed to the `main` branch.
    - [x] This PR is for a official card or any other directly to the integration related change and therefore directed to the `release` branch.

### DIFF
--- a/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_light.yaml
+++ b/custom_components/ui_lovelace_minimalist/lovelace/ulm_templates/card_templates/cards/card_light.yaml
@@ -275,32 +275,25 @@ card_light:
                 card:
                   - background-color: >
                       [[[
-                        if (!hass.themes.darkMode) {
-                          var color_set = variables.ulm_card_light_enable_color ? entity.attributes.rgb_color : variables.ulm_card_light_color;
-                          var color = 'rgba(var(--color-' + color_set + '),0.2)'
-                          if (variables.ulm_card_light_enable_color) {
+                        if (!hass.themes.darkMode && variables.ulm_card_light_force_background_color) {
+                          if (variables.ulm_card_light_enable_color && entity.state != "off") {
                             return 'rgba(250, 250, 250,0.8)';
                           }
-                          if (entity.state != "on") {
-                            return 'rgba(var(--color-theme),0.05)';
-                          }
-                          return color
                         }
                         return 'rgba(var(--color-theme),0.05)';
                       ]]]
                 icon:
                   - color: >
                       [[[
-                        if (!hass.themes.darkMode) {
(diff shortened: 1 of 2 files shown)


benbur98:
Nice work!

I like the white text, it's way more readable than the coloured text previously - I'll check some of the other cards because we might have some inconsistency between cards after this which wouldn't look too good...

For the background, I'm not sure if I prefer the new bold colour in light mode or the old pastel colour personally.
Could make it a variable but aware that the card might get overly complex...

Custom color choice is great too.

basbruss:
By looking at Behance the bold color background for any `active` state is definitely the way to go:
image
@schumijo want to you think on the design changes?

benbur98 (dismissed):
Tested - Card works well :)

benbur98:
image

What is everyones opinion on the colour of the buttons with this forced background? They are not very readable as is. Perhaps they should go white in this mode?

benbur98:
For the Button Colours, what is anyone's opinion on the preferred choice:

Option 1 - Change Icon Only
Option 2 - Set Button Background to White and Icon to the Bard Background Color - this is most similar to the Behance image above
Option 3 - Do Nothing...

image

basbruss:
@benbur98 Oh damn didn't test that feature haha.
Maybe give it the same values as the icon grid?  Same is done with the oscillating button from the card_fan

benbur98:
Sorry there was a bug in that code when not using the Collapse option, also some unnecessary parts.

Please change for:
```
                card:
                  - background-color: >
                      [[[
                        if (!hass.themes.darkMode && variables.mui_card_light_force_background_color) {
                          if (variables.mui_card_light_enable_color && entity.state != "off") {
                            return 'rgba(250, 250, 250,0.8)';
                          }
                        }
                        return 'rgba(var(--color-theme),0.05)';
                      ]]]
                icon:
                  - color: >
                      [[[
                        if (!hass.themes.darkMode && variables.mui_card_light_force_background_color) {
                          if (entity.state != "off") {
                            var color_set = variables.mui_card_light_enable_color ? entity.attributes.rgb_color : variables.mui_card_light_color;
                            var color = 'rgba(var(--color-' + color_set + '),1)';
                            if (variables.mui_card_light_enable_color) {
                              color = 'rgba(' + color_set + ',1)';
                            }
                            return color;
                          }
                        }
                        return 'rgba(var(--color-theme),0.9)';
                      ]]]
```

